PR #110, opened 2025-08-30
feat: Implement failover system for Gemini API keys

nirzaf:
This feature introduces a robust failover system by allowing users to add multiple Gemini API keys. The system automatically manages these keys, switching to a healthy one when the active key is rate-limited, thus ensuring service continuity.

- The configuration is updated to store an array of API keys.
- A `GeminiKeyManager` is implemented to handle key rotation and selection logic.
- The system automatically detects rate limit and invalid auth errors from the Gemini API and switches to the next available key.
- The settings UI is updated to allow users to add, view, and manage up to 5 Gemini API keys.
- Each key displays a visual status indicator (e.g., "Active", "Rate-Limited", or "Invalid").
- Users can manually switch the active key in the settings.

### DIFF
--- a/src/agent/gemini/index.ts
+++ b/src/agent/gemini/index.ts
@@ -5,6 +5,7 @@
  */
 
 // src/core/ConfigManager.ts
+import { GeminiKeyManager, KeyStatus } from '@/common/keyManager';
 import type { TModelWithConversation } from '@/common/storage';
 import { uuid } from '@/common/utils';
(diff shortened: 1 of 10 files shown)


IceyLiu:
@nirzaf Apologies, my friend. The multi-apikey solution differs somewhat from the form planned for AionUi. Approaches like limiting the configuration entry to Gemini or only modifying the CLI layer for support are not sufficiently universal. Therefore, we may need to reassess our approach. 
Currently, version 1.2.0 has introduced a simple apikey polling service, which involves modifications from the Core layer to the CLI layer. Welcome to discuss on the implementation and download it.

And, I would like to extend my most sincere respect. You did good job, looking forward to your next contribution!

PR #291, opened 2025-11-07
chore: compress .app to .zip before notarization upload to reduce tim…

kuishou68:


GemAION:
formal_verdict:
  verdict: BLOCK
  audited_by: bafuxunan
  role: 八府巡按 / Supervisor
  audit_target: PR #291
  audited_head: 96b5a05c55343468a26dc4c671eee75107c2ad41
  audit_url: https://github.com/iOfficeAI/AionUi/pull/291
  evidence_reviewed:
    - PR metadata: state=MERGED, base=dev, head=optimization/experience-improvement
    - Changed files: scripts/afterSign.js only
    - PR diff: compresses .app to .zip before notarytool submit, then deletes temporary ZIP
    - GitHub checks: no checks reported on the branch
    - Exact-head local validation: node --check scripts/afterSign.js
  findings:
    - BLOCKER: exact-head syntax validation fails before runtime because scripts/afterSign.js declares const fs twice inside the same function/block scope.
    - Evidence: node --check scripts/afterSign.js reports "SyntaxError: Identifier 'fs' has already been declared" at scripts/afterSign.js:84; the newly added declaration is at line 46 and the pre-existing declaration remains at line 84.
    - Scope note: changed path is limited to notarization/build scripting and does not touch application runtime, secrets, payment, DB, webhook, customer data, or real executor behavior.
    - State note: PR is already MERGED; this verdict is an audit write-back and does not merge, deploy, revert, or mutate production state.
  required_fixes_if_any:
    - Remove the duplicate fs declaration by hoisting a single const fs = require('fs') to module/function scope, or reuse the existing declaration before calling fs.unlinkSync.
    - Re-run syntax/build validation for scripts/afterSign.js after the fix.
  merge_allowed_after_verdict: false
  monarch_required: false
  reason:
    - A JavaScript syntax error in the changed build hook prevents the file from loading, so the notarization hook cannot execute reliably.
    - Missing CI/check coverage on the PR does not compensate for the failing exact-head local validation.

PR #751, opened 2026-02-08
feat: integrate Mistral Vibe CLI

### LINKED ISSUE #744: [Feature]: ACP integration with Mistral Vibe
### Feature Description

Addition of ACP integration with the CLI coding agent Vibe from Mistral https://github.com/mistralai/mistral-vibe

### Problem Statement

It expands on the available ACP integrations of AionUI, adding yet another option with Mistral Vibe

### Proposed Solution

Implement ACP integration for Mistral Vibe

### Feature Category

Multi-Agent Support

### Additional Context

_No response_

vthanhduong:
### Related issue
Closes #744 

### Solution
Integrated a new Acp Agent specifically for Mistral Vibe. This implementation allows the system to call Mistral Vibe CLI as a default provider, enhancing the multi-agent support.

### Changes
- **Agent Integration**: Added Mistral Vibe to the default AcpTypes.
- **Default Configuration**: Updated the provider list to include Vibe as a standard option for all users.

### Testing
- [x] Verified that the Mistral Vibe agent is correctly registered in the system.
- [x] Confirmed successful execution of commands via the Mistral Vibe CLI agent.
- [x] Tested alongside other agents to ensure multi-agent compatibility.


piorpua:
## Manual Review of AI-Generated Assessment & Code Review

I've verified each finding from the automated PR Assessment and Code Review against the actual source code on the PR branch. Here are the results:

### Summary

| # | Level | Issue | Verified? | Needs Fix? |
|---|-------|-------|-----------|------------|
| 1 | CRITICAL | `vibe` missing from `connect()` switch | **False positive** | No |
| 2 | CRITICAL | `acpArgs: ['']` passes empty string arg | **Confirmed** | Yes |
| 3 | HIGH | UI logo mapping not applied | **False positive** | No |

---

### CRITICAL #1: `vibe` backend not connectable — ❌ False Positive

The review claims `case 'vibe'` was added to "a different switch" (around line ~254) rather than the `connect()` method's switch, so `connect('vibe', ...)` would hit `default` and throw `Unsupported backend: vibe`.

**Actual finding**: There is only **one** switch on `backend` in the entire file (inside `connect()`). The PR correctly adds `case 'vibe':` to the generic CLI backend group in that switch. The bot was confused because the contributor's fork has different line numbers than `main`, leading it to incorrectly assume line ~254 was in a separate switch block.

**No action needed.**

---

### CRITICAL #2: `acpArgs: ['']` passes empty string argument — ✅ Confirmed

The current config:
```typescript
vibe: {
  // ...
  acpArgs: [''],
}
```

In `createGenericSpawnConfig()`, the logic is:
```typescript
const effectiveAcpArgs = acpArgs && acpArgs.length > 0 ? acpArgs : ['--experimental-acp'];
```

Since `[''].length > 0` evaluates to `true`, the empty string `''` will be used as an actual CLI argument instead of being treated as "no args." This will spawn something like `vibe-acp ""` which most CLIs would treat as an invalid/unknown argument.

**Suggested fix**: If `vibe-acp` doesn't need any ACP flags (which is likely, given the binary name already implies ACP mode), there are two options:

1. **Quick fix** — Change to an explicit no-op value and update `createGenericSpawnConfig` to distinguish `[]` (explicit no args) from `undefined` (use default fallback):
   ```typescript
   // In createGenericSpawnConfig:
   const effectiveAcpArgs = acpArgs === undefined ? ['--experimental-acp'] : acpArgs;
   
   // In acpTypes.ts:
   acpArgs: [],
   ```

2. **Minimal fix** — If you don't want to change `createGenericSpawnConfig`, you could set `acpArgs: undefined` (or omit it), and `vibe-acp` would receive `--experimental-acp` by default. Check whether `vibe-acp` accepts or ignores this flag.

---

### HIGH #3: UI logo mapping not applied — ❌ False Positive

The review claims the "full file content" doesn't include the logo changes. However, the bot itself noted: *"The following inputs were truncated due to size limits: file contents (>80K chars)"*. The truncated content led to a false conclusion.

The diff clearly shows all three files correctly add `import MistralLogo` and `vibe: MistralLogo`:
- `ChatLayout.tsx`
- `guid/index.tsx`
- `McpAgentStatusDisplay.tsx`

**No action needed.**

---

### Conclusion

Out of the 3 issues flagged, only **CRITICAL #2 (`acpArgs: ['']`)** is a real bug that needs to be fixed before merging. The other two findings are false positives caused by the review bot's inability to handle line number shifts (fork divergence) and truncated file content.

vthanhduong:
fixed it! pls check this PR again! thank you! @piorpua

piorpua:
Thank you @vthanhduong for this contribution! Great work integrating Mistral Vibe CLI support — this is a nice addition to our multi-agent ecosystem. 🎉

PR #773, opened 2026-02-09
fix(acp): prevent orphan CLI processes on conversation kill

YW1975:
## Problem

When a conversation using ACP agents (Claude, Codex, etc.) is killed — either by switching conversations, closing the app, or explicitly stopping — the CLI agent process tree is left running as orphans.

This happens because `AcpAgentManager` spawns CLI agents as child processes via `AcpConnection`, but the inherited `kill()` method only terminates the immediate worker without cleaning up the CLI process tree.

**Impact**: Over time, orphaned `claude`, `codex`, or other CLI processes accumulate, consuming system resources (CPU, memory, file handles).

## Solution

Override `kill()` in `AcpAgentManager` to:

1. Call `agent.stop()` first, which triggers `AcpConnection.disconnect()` → `treeKill(pid, 'SIGTERM')` to terminate the entire process tree
2. Add a 500ms grace period for treeKill's async pgrep + SIGTERM dispatch to complete
3. Include a 1.5s hard timeout fallback to prevent hanging if `stop()` gets stuck

```
kill() flow:
  ├─ Hard timeout (1.5s) → super.kill()    [safety net]
  └─ agent.stop()                           [graceful path]
       → AcpConnection.disconnect()
         → treeKill(pid, SIGTERM)
       → wait 500ms (grace period)
       → super.kill()
```

## Changes

- `src/process/task/AcpAgentManager.ts` — Added `kill()` override with graceful shutdown + hard timeout

## Testing

- Verified that CLI processes (claude, codex) are properly terminated when switching/closing conversations
- Verified hard timeout triggers correctly when `stop()` hangs

### DIFF
--- a/src/process/task/AcpAgentManager.ts
+++ b/src/process/task/AcpAgentManager.ts
@@ -363,6 +363,45 @@ class AcpAgentManager extends BaseAgentManager<AcpAgentManagerData, AcpPermissio
   }
 
   /**
+   * Override kill() to ensure ACP CLI process is terminated.
+   *
+   * Problem: AcpAgentManager spawns CLI agents (claude, codex, etc.) as child
+   * processes via AcpConnection. The default kill() from the base class only
+   * kills the immediate worker, leaving the CLI process running as an orphan.
+   *
+   * Solution: Call agent.stop() first, which triggers AcpConnection.disconnect()
+   * → ChildProcess.kill(). We add a grace period for the process to exit
+   * cleanly before calling super.kill() to tear down the worker.
+   *
+   * A hard timeout ensures we don't hang forever if stop() gets stuck.
+   * An idempotent doKill() guard prevents double super.kill() when the hard
+   * timeout and graceful path race against each other.
+   */
+  kill() {
+    let killed = false;
+    const GRACE_PERIOD_MS = 500; // Allow child process time to exit cleanly
+    const HARD_TIMEOUT_MS = 1500; // Force kill if stop() hangs
+
+    const doKill = () => {
+      if (killed) return;
+      killed = true;
+      clearTimeout(hardTimer);
+      super.kill();
+    };
+
+    // Hard fallback: force kill after timeout regardless
+    const hardTimer = setTimeout(doKill, HARD_TIMEOUT_MS);
+
+    // Graceful path: stop → grace period → kill
+    void (this.agent?.stop?.() || Promise.resolve())
+      .catch((err) => {
+        console.warn('[AcpAgentManager] agent.stop() failed during kill:', err);
+      })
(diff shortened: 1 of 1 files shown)


piorpua:
# Review Verification

I verified the Code Review and Assessment findings against the source code. Here are my conclusions:

## Code Review Issues Verification

### HIGH #1: `super.kill()` can be called twice (race condition) ✅ Confirmed, recommend fix

When `agent.stop()` + 500ms grace period exceeds the 1500ms hard timeout (plausible when CLI hangs or system is under load), `super.kill()` will be invoked by both the hardTimer callback and the `.finally()` block.

Currently `ForkTask.kill()` happens to be tolerant of this (`UtilityProcess.kill()` doesn't crash on already-dead processes, `process.off()` is idempotent), but relying on "happens to be safe" is fragile — if the base class ever adds event emission or state tracking, this breaks.

The suggested `killOnce` idempotent guard is the right fix:

```typescript
let killed = false;
const killOnce = () => {
  if (killed) return;
  killed = true;
  super.kill();
};
```

### HIGH #2: Swallowed `stop()` errors ✅ Confirmed, recommend fix

`.catch(() => {})` silently swallows all errors. Since the core goal of this PR is to prevent orphan processes, having no logs when cleanup fails will lead to "still leaks processes sometimes" reports with zero diagnostic information.

The project already has a consistent logging pattern (e.g., `console.error('[AcpAgentManager] Failed to save ACP session ID:', error)` at `AcpAgentManager.ts:427`). Suggest:

```typescript
.catch((err) => {
  console.warn('[AcpAgentManager] agent.stop() failed during kill()', err);
})
```

## Additional Finding: treeKill claim in PR description may be inaccurate

The PR description states the call chain triggers `treeKill(pid, 'SIGTERM')`, but `AcpConnection.disconnect()` actually uses standard `ChildProcess.kill()` (`AcpConnection.ts:956`), and no `tree-kill` dependency was found in the repo.

Standard `.kill()` only sends a signal to the **immediate child process**, not its descendants. If the CLI agent spawns sub-processes (LSP servers, tool subprocesses, etc.), those may still become orphans. Consider verifying this and either correcting the description or introducing `tree-kill` for actual process tree cleanup.

## Summary

| Issue | Confirmed | Recommendation |
|-------|-----------|----------------|
| Double kill race | ✅ | Fix before merge — add killOnce guard |
| Swallowed errors | ✅ | Fix before merge — add console.warn |
| treeKill description | ⚠️ | Verify actual implementation and correct description |

Both HIGH fixes are minimal (3-5 lines each) and should be addressed before merge.

YW1975:
Thanks for the thorough review @piorpua — all three confirmed. Fix pushed.

One note on #3: you're right, `disconnect()` uses standard `ChildProcess.kill()`, not treeKill. PR description updated accordingly.

— Ralph & Lisa

PR #825, opened 2026-02-11
fix(cron): improve scheduled task reliability and reuse agent connections

### LINKED ISSUE #824: Cron tasks miss executions during system sleep and reconnect every run
## Problem

Two reliability issues reported with scheduled/cron tasks:

### 1. Agent reconnects every execution
Every time a cron job fires, the existing agent connection is killed and recreated from scratch. This causes unnecessary delays and the user sees reconnection messages every execution cycle.

**Root cause:** `CronService.executeJob()` unconditionally calls `WorkerManage.kill()` on the existing task and creates a fresh instance with `yoloMode=true`, even when the existing task already has yoloMode enabled.

### 2. Tasks silently missed during system sleep
When the computer enters sleep/hibernate mode, all Node.js timers (`setInterval`, `setTimeout`, `croner`) are paused. After system wake:
- Timers may be in an invalid state
- Jobs that should have executed during sleep are silently skipped
- No notification is shown to the user
- No `powerMonitor` listener exists to detect sleep/wake transitions

**Impact:** Users who frequently open/close their laptops may miss scheduled task executions entirely without any indication.

## Expected Behavior

1. Cron jobs should reuse existing agent connections when possible
2. System sleep should not silently drop scheduled executions
3. Users should be notified when tasks are missed
4. The app should resist suspension while cron jobs are active

## Environment
- Platform: macOS / Windows / Linux
- Affects all agent types (ACP, Gemini, Codex, etc.)

kaizhou-lab:
## Summary

Fixes #824

- **Reuse agent connections**: Cron jobs now reuse existing agent instances via `ensureYoloMode()` instead of killing and recreating them every execution, eliminating unnecessary reconnections
- **PowerSaveBlocker**: Automatically enables `prevent-app-suspension` while cron jobs are active, keeping the app alive during system idle
- **Sleep/wake recovery**: Listens to Electron `powerMonitor.resume` event to detect missed jobs after system sleep, inserts warning messages into the affected conversations, and restarts all timers
- **Disable noisy messages**: Suppress `available_commands` messages in chat UI

### Agent connection reuse (fix reconnection)
- `BaseAgentManager`: Add `ensureYoloMode()` base method
- `AcpAgentManager`: Override with dynamic yoloMode enable via `session/set_mode` RPC
- `AcpAgent`: Add `enableYoloMode()` for runtime session mode switching
- `GeminiAgentManager`, `CodexAgentManager`, `OpenClawAgentManager`: Override to check existing yoloMode flag
- `CronService.executeJob()`: Try `ensureYoloMode()` first, only kill+recreate as fallback

### Sleep/wake reliability
- `CronService`: Add `powerSaveBlocker` management (auto start/stop based on active jobs)
- `CronService.handleSystemResume()`: Detect missed jobs, insert warning messages into conversations, restart timers
- `CronStore`/`ipcBridge`: Add `'missed'` status type
- `index.ts`: Register `powerMonitor.on('resume')` listener

### UI cleanup
- Disable `available_commands` messages (AcpAdapter, chatLib, MessageList)

## Test plan

- [ ] Create a cron job, verify it executes without reconnecting each time
- [ ] Sleep the computer past a scheduled execution time, wake and verify:
  - Warning message appears in the conversation
  - Job state shows `missed` status
  - Timer restarts and next execution works normally
- [ ] Enable/disable cron jobs and verify powerSaveBlocker starts/stops accordingly
- [ ] Verify all agent types (ACP, Gemini, Codex) work with the new ensureYoloMode flow


kaizhou-lab:
截屏2026-02-11 下午9 24 15

PR #831, opened 2026-02-12
fix(katex): improve math rendering and support latex code blocks

### LINKED ISSUE #789: [Bug]: LaTeX/Math expressions render twice - once as formatted equation and once as raw code
### Platform

macOS (Apple Silicon)

### AionUi Version

1.8.4

### Bug Description

When the LLM generates responses containing mathematical expressions (both inline and display) using LaTeX syntax, each expression appears twice in the chat interface:

1. Once as a properly rendered mathematical expression
2. Once as the raw LaTeX code alongside it

This creates a duplicate, side-by-side display (please see attached screenshot) that clutters the interface and makes it difficult to read mathematical content.

### Steps to Reproduce

1. Start a conversation with any LLM through AionUI
2. Instruct the LLM to format inline math using `$...$` and display math using `$$...$$` (without additional spaces inside the delimiters)
3. Ask the LLM to generate a response containing mathematical expressions
4. Observe that each math expression appears twice: rendered + raw code

### Expected Behavior

Mathematical expressions should render only once as properly formatted equations, with the raw LaTeX code hidden from view.

### Actual Behavior

Each mathematical expression appears twice - the rendered version appears alongside the raw LaTeX source code.

### Additional Context

This appears to be a markdown/LaTeX rendering issue in the chat interface. Similar issues have been reported in other projects:

- Jan AI: https://github.com/janhq/jan/issues/2865
- Astro Theme Typography: https://github.com/moeyua/astro-theme-typography/issues/176

The issue may be related to how the markdown parser handles LaTeX delimiters (`$...$` for inline and `$$...$$` for display math).

Image

kaizhou-lab:
## Summary

Fixes #789

- **Fix vertical overlap in fractions/superscripts**: Override global `* { line-height: 26px }` in Shadow DOM with `.katex, .katex * { line-height: normal }` so KaTeX can use its own line-height for proper rendering
- **Remove forced scrollable box on inline math**: Previously all `.katex` spans had `overflow-x: auto` + `display: inline-block` applied, causing every formula to appear in a scrollable container. Now only `.katex-display` (block math) gets horizontal overflow
- **Render `latex`/`math`/`tex` code blocks as math**: LLMs often output math formulas in ` ```latex ` code blocks instead of `$...$` delimiters. These are now rendered as KaTeX display math
- **Skip full LaTeX documents**: Code blocks containing `\documentclass`, `\usepackage`, or `\begin{document}` are detected as full LaTeX documents and rendered as syntax-highlighted code blocks instead of attempting KaTeX parsing (which would produce red error text)

## Files Changed

- `src/renderer/components/Markdown.tsx` — Shadow DOM CSS fixes, span handler cleanup, latex code block rendering
- `src/renderer/pages/conversation/preview/components/viewers/MarkdownViewer.tsx` — latex code block rendering for Streamdown preview

## Test Plan

- [ ] Inline math (`$x^2$`) renders without scrollable box
- [ ] Block math (`$$\frac{a}{b}$$`) renders with proper fraction layout (no vertical overlap)
- [ ] ` ```latex ` code block with math formula renders as KaTeX display math
- [ ] ` ```latex ` code block with full LaTeX document (`\documentclass{article}...`) renders as syntax-highlighted code
- [ ] Dark/light theme switching works correctly for all math rendering

### DIFF
--- a/src/renderer/components/Markdown.tsx
+++ b/src/renderer/components/Markdown.tsx
@@ -13,6 +13,7 @@ import remarkBreaks from 'remark-breaks';
 import remarkGfm from 'remark-gfm';
 import remarkMath from 'remark-math';
 
+import katex from 'katex';
 // Import KaTeX CSS to make it available in the document
 import 'katex/dist/katex.min.css';
 
@@ -92,6 +93,25 @@ function CodeBlock(props: any) {
     const match = /language-(\w+)/.exec(className || '');
     const language = match?.[1] || 'text';
     const codeTheme = currentTheme === 'dark' ? vs2015 : vs;
+
+    // Render latex/math code blocks as KaTeX display math
+    // Skip full LaTeX documents (with \documentclass, \begin{document}, etc.) — KaTeX only handles math
+    if (language === 'latex' || language === 'math' || language === 'tex') {
+      const latexSource = String(children).replace(/\n$/, '');
+      const isFullDocument = /\\(documentclass|begin\{document\}|usepackage)\b/.test(latexSource);
+      if (!isFullDocument) {
+        try {
+          const html = katex.renderToString(latexSource, {
+            displayMode: true,
+            throwOnError: false,
+          });
+          return <div className='katex-display' dangerouslySetInnerHTML={{ __html: html }} />;
+        } catch {
+          // Fall through to render as code block if KaTeX fails
+        }
+      }
+    }
+
     if (!String(children).includes('\n')) {
       return (
         <code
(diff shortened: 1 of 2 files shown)


piorpua (approved):

ParthoM7:
Thanks for the update, and sorry for the delay in responding. I've now updated the app via `git checkout main` and tested it further. 

**Current Observation**

Although the boxes are not present anymore and the rendering is smooth in that respect, the overlapping problem is still present. You may please consult the screenshot attached. The $a_{n+2}$ equation was written using double-dollar:

`$$ a_{n+2} = \frac{-\left[ (n+L+1)(n+L+2) - \ell(\ell+1) \right]}{\frac{(n+1)(n+2)}{ \left( \frac{2\mu E}{\hbar^2} \right) + \frac{k^2}{n^2}}} a_n $$`

which should normally be rendered as,

$$ a_{n+2} = \frac{-\left[ (n+L+1)(n+L+2) - \ell(\ell+1) \right]}{\frac{(n+1)(n+2)}{ \left( \frac{2\mu E}{\hbar^2} \right) + \frac{k^2}{n^2}}} a_n $$

As seen in the attachment, in this case the effect is severe. However, mild effects are still visible for simpler fractions as well. 

Screenshot 2026-02-14 at 4 47 07 PM

PR #862, opened 2026-02-13
fix(webui): prevent white screen when navigating to parent directory

### LINKED ISSUE #849: [Bug]: 安卓手机 打开 webui，在选择文件后页面变白
### Platform

Windows

### AionUi Version

1.8.7

### Bug Description

Image

### Steps to Reproduce

打开 pc端的 web ui，
使用安卓手机连接
选择文件，返回上一页
画面变白

### Expected Behavior

应该能返回上一页

### Actual Behavior

画面变白

### Additional Context

_No response_

kaizhou-lab:
## Summary

Fixes #849

- **Backend**: Add `isPathAllowed` helper and fix `canGoUp` logic in `directoryApi.ts` to verify the parent directory is within the allowed scope before exposing the "go up" option
- **Frontend**: Add defensive checks in `DirectorySelectionModal.tsx` to handle failed API responses gracefully, preserving the current view instead of crashing

## Root Cause

When browsing directories on Android WebUI, the `canGoUp` flag only checked whether the current directory was the filesystem root (`/`). It did not check whether the parent directory was within the allowed access scope (`cwd` and `homedir`). Clicking ".." from the home directory sent a request to `/Users`, which returned a 500 error. The frontend set the error response (no `items` field) as state, causing `undefined.map()` → TypeError → React white screen.

## Test Plan

- [ ] Open WebUI on mobile browser, navigate to home directory → ".." button should NOT appear
- [ ] Navigate into subdirectories → ".." button works normally to go back
- [ ] At `cwd` or `homedir` top level, ".." only shows if parent is also in allowed scope
- [ ] Desktop Electron behavior unchanged

### DIFF
--- a/src/renderer/components/DirectorySelectionModal.tsx
+++ b/src/renderer/components/DirectorySelectionModal.tsx
@@ -46,12 +46,10 @@ const DirectorySelectionModal: React.FC<DirectorySelectionModalProps> = ({ visib
           credentials: 'include',
         });
         if (!response.ok) {
-          console.error('Directory browse failed:', response.status);
           return;
         }
         const data = await response.json();
         if (!data || !Array.isArray(data.items)) {
-          console.error('Invalid directory response:', data);
           return;
         }
         setDirectoryData(data);


kaizhou-lab:
Both issues from the code review have been addressed:

1. **HIGH — `console.error`**: Removed the newly added `console.error` calls. The pre-existing one in the `catch` block is unchanged. On error, we now silently preserve the current directory view. → 6e0e3a3b

2. **MEDIUM — symlink canonicalization**: Added `fs.realpathSync` in `isPathAllowed` to stay consistent with `validatePath` and `DEFAULT_ALLOWED_DIRECTORIES`. → eb3f8297

PR #926, opened 2026-02-20
fix: make Codex CLI the single writer for apply_patch

seonhwi07jp:
## Summary
This PR intentionally includes only one scope: apply_patch write-path behavior.

- Remove local patch application in CodexAgentManager approval flows.
- Keep Codex CLI as the only writer after approval.
- No parser or unified-diff algorithm change.

## Why This Change Was Needed
During approval handling, AionUI could apply file edits locally before or in parallel with Codex CLI applying the same patch. That creates a dual-writer path and can desynchronize:

- engine return state vs actual file state
- verification result vs on-disk result
- especially under encoding/EOL-sensitive diffs

This is why users can observe failed result but file changed class symptoms.

## Root Cause (Code Evidence)
The same patch could be written twice from two independent paths:

1. Manual confirmation path
- local decision is stored, then approval is sent to CLI (respondElicitation).
- previously this flow also applied local patch changes before approval send.
- evidence: src/process/task/CodexAgentManager.ts:395, src/process/task/CodexAgentManager.ts:411

2. Auto-approve path
- previously also executed local patch apply in fire-and-forget mode, then sent approval.
- evidence: src/process/task/CodexAgentManager.ts:636, src/process/task/CodexAgentManager.ts:640

## How Gemini and Other Local Backends Handle This
In existing non-Codex managers, approval is forwarded and execution happens in the agent/worker path; manager layer does not perform a second local write.

- Gemini manager forwards confirmation via worker message only:
  - src/process/task/GeminiAgentManager.ts:656
  - src/process/task/GeminiAgentManager.ts:670
- Gemini tool scheduler separates `awaiting_approval` and `executing` states in one pipeline:
  - src/agent/gemini/cli/useReactToolScheduler.ts:125
  - src/agent/gemini/cli/useReactToolScheduler.ts:194
- ACP manager forwards confirmation to agent:
  - src/process/task/AcpAgentManager.ts:446
  - src/process/task/AcpAgentManager.ts:449
- OpenClaw manager forwards confirmation to agent:
  - src/process/task/OpenClawAgentManager.ts:204
  - src/process/task/OpenClawAgentManager.ts:209

This PR aligns Codex manager behavior with that single-writer model.

## What Changed
- Removed local applyPatchChanges(...) execution from:
  - explicit confirmation path
  - auto-approve path
- Added explicit guard comments that Codex CLI is the single writer.

## Scope / Risk
- Scope is minimal: 1 file (src/process/task/CodexAgentManager.ts), 1 commit.
- Low regression risk for unrelated features.
- Behavior change is targeted: write responsibility is now centralized in CLI.

## Validation
- Branch diff against upstream/main contains only:
  - src/process/task/CodexAgentManager.ts
- Commit set against upstream/main contains only:
  - 787d1861 (fix: make codex CLI the single writer for apply_patch)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "AionUi",
-  "version": "1.8.14",
+  "version": "1.8.15",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "AionUi",
-      "version": "1.8.14",
+      "version": "1.8.15",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "AionUi",
   "productName": "AionUi",
-  "version": "1.8.14",
+  "version": "1.8.15",
   "description": "Transform your command-line AI agent into a modern, efficient AI Chat interface.",
   "main": ".webpack/main",
   "scripts": {


piorpua:
Thanks for the fix @seonhwi07jp! This correctly makes Codex CLI the single writer for `apply_patch`, eliminating the double-write issue.

I've removed the now-dead `applyPatchChanges` method in #928 as a follow-up cleanup.

PR #980, opened 2026-02-25
fix(webui): resolve packaged renderer path lookup

kuishou68:
## Summary
- fix renderer asset path resolution for packaged Windows app
- support both electron-vite output path and legacy webpack fallback
- improve error output with all attempted renderer paths for easier diagnostics

## User-facing fix
Resolves startup failure after packaging where app reported:
" Renderer assets not found at ...app.asar/.webpack/renderer/main_window/index.html\


ringringlin (approved):

Moneyman011:
Greetings to you 👋

PR #1195, opened 2026-03-06
fix: normalize electron-updater metadata to ensure deterministic release builds

loonghao:
## Problem

When building releases for multiple architectures (x64/arm64) on multiple platforms (Windows/macOS/Linux), each build job produces updater metadata files with the same name (e.g., `latest.yml`, `latest-mac.yml`). When these artifacts are downloaded and collected for the release, the order in which files are copied is **nondeterministic** — whichever job's file happens to be processed last wins.

This causes:
- **Incorrect auto-update metadata**: e.g., `latest.yml` might contain arm64 metadata instead of x64, causing users on the wrong architecture to receive broken updates
- **Flaky releases**: The same CI pipeline can produce different release metadata on each run
- **Hard-to-debug update failures**: Users report update issues that can't be reproduced because the metadata changes between builds

## Solution

Added a new **"Prepare release assets (normalize updater metadata)"** step in the release job that:

1. **Copies all distributables** (`.exe`, `.msi`, `.dmg`, `.deb`, `.AppImage`, `.zip`, `.blockmap`) into a flat `release-assets/` directory
2. **Explicitly selects canonical metadata** from the correct platform-specific artifact directory:
   - `latest.yml` ← from `windows-build-x64`
   - `latest-mac.yml` ← from `macos-build-x64`
   - `latest-linux.yml` ← from `linux-build` (x64)
   - `latest-linux-arm64.yml` ← from `linux-build` (arm64)
3. **Publishes architecture-scoped metadata** for diagnostics (`latest-win-x64.yml`, `latest-win-arm64.yml`, `latest-mac-x64.yml`, `latest-mac-arm64.yml`)
4. **Publishes architecture-scoped debug metadata** (`builder-debug-win-x64.yml`, etc.)
5. **Validates required metadata** — fails the build if any required updater file is missing

The release step now uploads from `release-assets/` instead of `build-artifacts/`, ensuring **deterministic, correct metadata** every time.

## Changes

- `.github/workflows/build-and-release.yml`: Added metadata normalization step in the `release` job

### DIFF
--- a/scripts/create-mock-release-artifacts.sh
+++ b/scripts/create-mock-release-artifacts.sh
@@ -67,6 +67,7 @@ echo "debug: mac-arm64" > "$ARTIFACTS_DIR/macos-build-arm64/builder-debug.yml"
 
 # Linux
 touch "$ARTIFACTS_DIR/linux-build/AionUi-1.0.0.AppImage"
+touch "$ARTIFACTS_DIR/linux-build/AionUi-1.0.0-arm64.AppImage"
 touch "$ARTIFACTS_DIR/linux-build/AionUi-1.0.0.deb"
 cat > "$ARTIFACTS_DIR/linux-build/latest-linux.yml" <<'EOF'
 version: 1.0.0
--- a/scripts/prepare-release-assets.sh
+++ b/scripts/prepare-release-assets.sh
@@ -16,13 +16,14 @@ set -euo pipefail
 ARTIFACTS_DIR="${1:-build-artifacts}"
 OUTPUT_DIR="${2:-release-assets}"
 
+rm -rf "$OUTPUT_DIR"
 mkdir -p "$OUTPUT_DIR"
 
 # ---------------------------------------------------------------------------
 # 1) Copy all distributables and blockmaps (unique file names)
 # ---------------------------------------------------------------------------
 echo "==> Copying distributables from $ARTIFACTS_DIR ..."
-find "$ARTIFACTS_DIR" -type f \( \
+mapfile -t DISTRIBUTABLES < <(find "$ARTIFACTS_DIR" -type f \( \
   -name "*.exe" -o \
   -name "*.msi" -o \
   -name "*.dmg" -o \
@@ -30,7 +31,18 @@ find "$ARTIFACTS_DIR" -type f \( \
   -name "*.AppImage" -o \
   -name "*.zip" -o \
   -name "*.blockmap" \
-\) -exec cp -f "{}" "$OUTPUT_DIR/" \;
+\) | sort)
+
+DUPLICATE_BASENAMES=$(for file in "${DISTRIBUTABLES[@]}"; do basename "$file"; done | sort | uniq -d || true)
+if [ -n "$DUPLICATE_BASENAMES" ]; then
+  echo "::error::Found duplicate distributable basenames that would be overwritten in flat output:"
+  echo "$DUPLICATE_BASENAMES"
+  exit 1
+fi
+
+for file in "${DISTRIBUTABLES[@]}"; do
+  cp -f "$file" "$OUTPUT_DIR/"
+done
 
 # ---------------------------------------------------------------------------
 # 2) Collect updater metadata from each platform artifact directory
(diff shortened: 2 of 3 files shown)


piorpua:
# Review Verification

I verified the Code Review findings against the source code and PR diff.

## Code Review Issues Verification

### CRITICAL #1: Release workflow calls a script that isn't present — ❌ False Positive

The AI review claimed there's a mismatch — that the workflow "still uploads from `build-artifacts/**`" and "does not include" the new Prepare release assets step.

**Actual finding**: The PR diff is internally consistent and correct. It adds **both**:
1. The new `Prepare release assets (normalize updater metadata)` step calling `bash scripts/prepare-release-assets.sh build-artifacts release-assets`
2. Updates the `Create Release` files glob from `build-artifacts/**` to `release-assets/**`

Additionally, `scripts/prepare-release-assets.sh` is committed as a **new file in this very PR**. The workflow calls it via `bash scripts/...` (not as an executable), so no executable bit is required.

The AI reviewer appears to have confused the base branch state (which still has the old `build-artifacts/**` glob) with the PR's proposed changes. This is a false positive.

---

### HIGH #2: Linux metadata find not constrained to `linux-build/` — ✅ Confirmed

Looking at `scripts/prepare-release-assets.sh` lines 37–38:

```bash
LINUX_X64_LATEST=$(find "$ARTIFACTS_DIR" -type f -name "latest-linux.yml" | sort | head -n 1 || true)
LINUX_ARM64_LATEST=$(find "$ARTIFACTS_DIR" -type f -name "latest-linux-arm64.yml" | sort | head -n 1 || true)
```

Unlike Windows and macOS, Linux finds don't use a `-path "*/linux-build/*"` constraint. Looking at the actual workflow matrix, there is only one `linux-build` artifact (the Linux job produces both x64 and arm64 artifacts in a single job with `"arch":"x64-arm64"`), so the practical risk of picking the wrong file is low. However, the inconsistency is real — if another artifact directory ever contained a `latest-linux.yml`, the script would non-deterministically pick the wrong one. The fix is trivial:

```bash
LINUX_X64_LATEST=$(find "$ARTIFACTS_DIR" -type f -path "*/linux-build/*" -name "latest-linux.yml" | sort | head -n 1 || true)
LINUX_ARM64_LATEST=$(find "$ARTIFACTS_DIR" -type f -path "*/linux-build/*" -name "latest-linux-arm64.yml" | sort | head -n 1 || true)
```

---

### HIGH #3: Flat copy of distributables can silently overwrite duplicate basenames — ⚠️ Minor

The concern about silent overwrite via `cp -f` is real in theory. However, electron-builder generates architecture-specific filenames by convention (e.g., `AionUi-1.8.23-arm64.dmg` vs `AionUi-1.8.23.dmg`), so collisions are very unlikely in practice with standard electron-builder configuration. The suggestion to add collision detection is reasonable defensive programming, but the practical risk does not warrant a HIGH severity rating for this project.

---

### MEDIUM #4: `verify-release-assets.sh` uses mock-specific strings — ✅ Confirmed

The script hardcodes `fake-sha512-x64`, `AionUi-1.0.0-win-x64.exe`, etc. This is fine as a CI unit test for the normalization logic, but it won't catch real-world metadata mismatches (e.g., `latest.yml` referencing an arm64 path instead of x64). Adding structural checks — such as verifying that `latest.yml` contains a `path:` field ending with a recognizable x64 pattern, and that the referenced file actually exists in `release-assets/` — would increase confidence in production releases.

---

## Additional Findings

No additional issues missed. The overall approach is sound: the root cause (nondeterministic artifact merging) is correctly identified and the solution (explicit per-platform metadata selection) addresses it effectively. No security vulnerabilities, logic errors, or immutability violations found.

## Summary

| # | Severity | Issue | Verdict | Recommendation |
|---|----------|-------|---------|----------------|
| 1 | CRITICAL | Workflow missing normalization step / script not committed | ❌ False Positive | No action needed |
| 2 | HIGH | Linux metadata find not constrained to `linux-build/` | ✅ Confirmed | Add `-path "*/linux-build/*"` constraint |
| 3 | HIGH | Flat copy may silently overwrite duplicate basenames | ⚠️ Minor | Low practical risk; optional collision detection |
| 4 | MEDIUM | Verify script uses mock-specific strings | ✅ Confirmed | Consider adding structural content checks |

## Conclusion

This PR solves a real and important problem — nondeterministic release metadata in multi-arch CI builds. The implementation is correct. Issue #2 is a one-line fix worth applying for consistency and future-proofing. Issues #3 and #4 are optional improvements. The CRITICAL finding from the AI review was a false positive.

The PR is suitable for merge after addressing Issue #2.

PR #1196, opened 2026-03-06
fix(conversation): refresh sidebar on AI response finish

### LINKED ISSUE #1160: [Bug]: 对话列表没有实时刷新
### Platform

macOS (Apple Silicon)

### AionUi Version

1.8.20

### Bug Description

Image

### Steps to Reproduce

n/a

### Expected Behavior

n/a

### Actual Behavior

n/a

### Additional Context

_No response_

audichuang:
## Problem

Fixes #1160

The conversation list (sidebar) does not refresh in real-time after AI finishes responding. As noted by @IceyLiu: "last modify被搞掉了" — the `modifyTime` update was not triggering a sidebar re-sort.

## Root Cause

When AI agents finish responding, the backend correctly updates `modifyTime` via `updateConversation()`, but the frontend's `finish` event handler only resets loading states — it does **not** emit `chat.history.refresh` to notify the sidebar to re-fetch and re-sort.

This means:
- Conversations don't move to the top of the list after receiving AI responses
- Cron-triggered conversations don't update their position
- Users must manually refresh to see correct sorting

## Fix

Add `emitter.emit('chat.history.refresh')` in the `finish` event handler of all 5 agent SendBox components:

- `AcpSendBox.tsx` (Claude, Qwen, etc.)
- `GeminiSendBox.tsx`
- `CodexSendBox.tsx`
- `OpenClawSendBox.tsx`
- `NanobotSendBox.tsx`

## Changes

5 files changed, 10 insertions (each file gets 2 lines: a comment + the emit call).


kaizhou-lab:
Thanks for the investigation and the clean PR!

However, this change won't actually fix the sidebar sorting issue. Here's why:

### How sidebar sorting works

The sidebar sorts conversations by `modifyTime`. When the frontend emits `chat.history.refresh`, the sidebar re-fetches conversations from the database and sorts them by `modifyTime` (see `databaseBridge.ts:60`, `ChatHistory.tsx:108`).

### What this PR does

This PR adds `emitter.emit('chat.history.refresh')` in the `finish` event handler. But the sidebar **already refreshes** when the user sends a message — every `onSendHandler` already emits `chat.history.refresh`.

### The real problem

The `modifyTime` is only updated in the database when:
- A user message is sent (`GeminiAgentManager:329`, `AcpAgentManager:546` — calls `updateConversation({})`)
- Configuration changes (sessionMode, modelId, etc.)

**No AgentManager calls `updateConversation` when AI finishes responding** (`this.status = 'finished'`). So even if the frontend emits `chat.history.refresh` again at finish time, the `modifyTime` in the database is still the same value from when the user sent the message — the sort order doesn't change.

### The actual fix

The fix should be in the **backend** AgentManagers — call `getDatabase().updateConversation(this.conversation_id, {})` when the AI response finishes, so `modifyTime` gets updated to the completion time. The existing frontend `chat.history.refresh` emit (from `onSendHandler`) would then already pick up the correct sorting.

Closing this PR since the frontend-only approach doesn't achieve the intended effect. We'll address this from the backend side.

PR #1216, opened 2026-03-09
fix(cron): fix unread red dot indicator not disappearing on conversation click

cdxiaodong:
## Summary
Fixed two critical issues with the cron job unread indicator (red dot):
1. **Electron**: Red dot remained visible after clicking a conversation with unread cron executions
2. **WebUI**: Red dot was not displayed when cron jobs executed

## Root Cause
Multiple components were calling `useCronJobsMap()` independently, creating separate state instances:
- `ConversationRow.tsx` - displayed red dots based on its own instance
- `WorkspaceGroupedHistory` (index.tsx) - called markAsRead on its own instance
- These were different instances, so state wasn't shared between display and interaction

## Changes
- Remove `useCronJobsMap()` call from `ConversationRow` component
- Pass `getJobStatus` as a prop from parent `WorkspaceGroupedHistory`
- Add `useEffect` in `WorkspaceGroupedHistory` to call `markAsRead` when conversation id changes
- Add `unreadConversations` to `useMemo` dependencies to trigger re-renders
- Ensure single source of truth for `unreadConversations` state across all components

## Test Plan
- [x] **Electron**: Create cron job → Wait for execution (red dot appears) → Click conversation → Verify red dot disappears
- [x] **WebUI**: Create cron job → Wait for execution → Verify red dot appears in conversation list → Click conversation → Verify red dot disappears
- [x] Both platforms now use the same state instance, ensuring consistent behavior

## Technical Details
The fix ensures:
1. When a cron job executes, `onJobUpdated` event updates the single `unreadConversations` Set
2. Red dots appear in both Electron and WebUI based on this shared state
3. When user clicks conversation, `markAsRead(id)` is called on the same state instance
4. `useMemo` detects `unreadConversations` change and triggers re-render
5. Red dot disappears immediately in both platforms ✅

### DIFF
--- a/src/renderer/pages/cron/hooks/useCronJobs.ts
+++ b/src/renderer/pages/cron/hooks/useCronJobs.ts
@@ -254,8 +254,6 @@ export function useCronJobsMap() {
   const lastRunAtMapRef = useRef<Map<string, number>>(new Map());
   // Track current active conversation (use ref to access latest value in event handlers)
   const activeConversationIdRef = useRef<string | null>(null);
-  // Force update trigger for re-rendering
-  const [trigger, setTrigger] = useState(0);
 
   // Persist unread state to localStorage
   useEffect(() => {
@@ -429,6 +427,12 @@ export function useCronJobsMap() {
     });
   }, []);
 
+  // Update active conversation ref without triggering state update
+  // Use this to sync the ref when route changes (e.g., URL navigation)
+  const setActiveConversation = useCallback((conversationId: string) => {
+    activeConversationIdRef.current = conversationId;
+  }, []);
+
   // Check if a conversation has unread cron executions
   const hasUnread = useCallback(
     (conversationId: string) => {
@@ -445,10 +449,11 @@ export function useCronJobsMap() {
       getJobsForConversation,
       getJobStatus,
       markAsRead,
+      setActiveConversation,
       hasUnread,
       refetch: fetchAllJobs,
     }),
-    [jobsMap, loading, hasJobsForConversation, getJobsForConversation, getJobStatus, markAsRead, hasUnread, fetchAllJobs, unreadConversations]
+    [jobsMap, loading, hasJobsForConversation, getJobsForConversation, getJobStatus, markAsRead, setActiveConversation, hasUnread, fetchAllJobs]
   );
 }
 
--- a/src/renderer/pages/conversation/grouped-history/hooks/useConversationActions.ts
+++ b/src/renderer/pages/conversation/grouped-history/hooks/useConversationActions.ts
@@ -15,7 +15,6 @@ import { useNavigate, useParams } from 'react-router-dom';
 
 import { useConversationTabs } from '../../context/ConversationTabsContext';
 import { isConversationPinned } from '../utils/groupingHelpers';
-import { useCronJobsMap } from '@/renderer/pages/cron';
 
 type UseConversationActionsParams = {
   batchMode: boolean;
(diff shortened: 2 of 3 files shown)


src/renderer/pages/cron/hooks/useCronJobs.ts
```diff
@@ -254,6 +254,8 @@ export function useCronJobsMap() {
   const lastRunAtMapRef = useRef<Map<string, number>>(new Map());
   // Track current active conversation (use ref to access latest value in event handlers)
   const activeConversationIdRef = useRef<string | null>(null);
+  // Force update trigger for re-rendering
+  const [trigger, setTrigger] = useState(0);
```
kaizhou-lab: unused code ，please remove

kaizhou-lab:
## Memory Leak & Performance Review

### 核心问题：`useCronJobsMap()` 被重复实例化

`useCronJobsMap()` 不是基于 Context 的共享 hook，**每次调用都会创建一套完全独立的状态和 IPC 事件订阅**。PR 修改后，在 `WorkspaceGroupedHistory` 的组件树内存在 **2 个独立实例**：

| Location | File | Line |
|------|------|------|
| Instance 1 | `grouped-history/index.tsx` | :29 |
| Instance 2 | `grouped-history/hooks/useConversationActions.ts` | :39 |

Each instance independently:
- Creates its own `jobsMap`, `unreadConversations`, `loading` state
- Registers **3 IPC event listeners** (`onJobCreated`, `onJobUpdated`, `onJobRemoved`)
- Calls `fetchAllJobs()` on mount (IPC call)
- Persists to `localStorage`

### Issue 1: Cascading render storm on conversation click

When user clicks a conversation item:

1. `handleConversationClick` → calls **Instance 2**'s `markAsRead(id)` → triggers `setUnreadConversations` → **render #1**
2. Navigation happens, `id` changes via `useParams()`
3. `index.tsx` `useEffect([id, markAsRead])` → calls **Instance 1**'s `markAsRead(id)` → triggers `setUnreadConversations` → **render #2**

Two independent instances each trigger a full render cycle on every click.

### Issue 2: Double event handling on cron execution

When a cron job executes, `onJobUpdated` IPC event is handled by **both instances**:

- Instance 1's `onJobUpdated` → `setUnreadConversations` + `emitter.emit('chat.history.refresh')`
- Instance 2's `onJobUpdated` → `setUnreadConversations` + `emitter.emit('chat.history.refresh')`

`chat.history.refresh` fires **twice**, causing the conversation list to be fetched and re-rendered **twice**.

### Issue 3: IPC listener accumulation

2 instances × 3 IPC listeners = **6 event listeners** (should be 3).

### Issue 4: Redundant `useMemo` dependency

```ts
// useCronJobs.ts:451
[jobsMap, loading, ..., getJobStatus, markAsRead, hasUnread, fetchAllJobs, unreadConversations]
//                                                                         ^^^^^^^^^^^^^^^^^^^^
```

`unreadConversations` is redundant here — `getJobStatus` and `hasUnread` already depend on it via their `useCallback` deps. Adding it to the final `useMemo` causes **unnecessary return object recreation**, triggering re-renders in all consumers.

### Issue 5: Dead code

```ts
// useCronJobs.ts:258
const [trigger, setTrigger] = useState(0);  // declared but setTrigger is never called
```

### Suggested Fix

**Best approach**: Convert `useCronJobsMap` to a React Context Provider so there's only one instance app-wide.

**Minimum fix**:
1. Remove `useCronJobsMap()` from `useConversationActions` — pass `markAsRead` as a parameter instead
2. Remove the duplicate `markAsRead` `useEffect` in `index.tsx` (since `handleConversationClick` already calls it)
3. Remove redundant `unreadConversations` from `useMemo` deps
4. Remove unused `trigger` state

cdxiaodong:
@kaizhou-lab  已优化

PR #1223, opened 2026-03-09
fix(telegram): convert Markdown to HTML for proper message rendering

### LINKED ISSUE #1214: [Bug]: Telegram bot does not render Markdown (shows raw asterisks)
### Platform

Linux

### AionUi Version

1.18.23

### Bug Description

The Telegram bot integration currently fails to render Markdown formatting (like **bold**, *italic*, code blocks) correctly. Instead of formatted text, it displays the raw Markdown syntax (e.g., literal asterisks).

Root Cause:
The Telegram plugin hardcodes parse_mode: "HTML" (likely because system prompts/confirmations use HTML tags). However, the formatTextForPlatform function currently treats Telegram the same as generic text, running only escapeHtml() on the content. When an LLM outputs Markdown, it is escaped and sent as-is, which Telegram displays as plain text.

Proposed Fix:
I have patched this locally by modifying the formatTextForPlatform function to include a specific case for Telegram that converts basic Markdown to Telegram-compatible HTML tags.

Patch Logic (JavaScript):
In out/main/index.js (or the equivalent TypeScript source), update formatTextForPlatform to handle the telegram platform:

JavaScript


function formatTextForPlatform(text, platform) {
  if (platform === "lark") {
    return convertHtmlToLarkMarkdown(text);
  }
  if (platform === "dingtalk") {
    return convertHtmlToDingTalkMarkdown(text);
  }
  
  // FIX: Convert Markdown to HTML for Telegram
  if (platform === "telegram") {
    let formatted = escapeHtml(text);
    // Bold
    formatted = formatted.replace(/\*\*(.*?)\*\*/g, "<b>$1</b>");
    // Italic
    formatted = formatted.replace(/\*(.*?)\*/g, "<i>$1</i>");
    // Code blocks
    formatted = formatted.replace(/```([\s\S]*?)```/g, "<pre><code>$1</code></pre>");
    // Inline code
    formatted = formatted.replace(/`([^`]+)`/g, "<code>$1</code>");
    // Links
    formatted = formatted.replace(/\[([^\]]+)\]\(([^)]+)\)/g, "<a href=\"$2\">$1</a>");
    // Lists (optional improvement)
    formatted = formatted.replace(/^\s*-\s+/gm, "• ");
    
    return formatted;
  }

  return escapeHtml(text);
}

Environment:

• AionUi Version: (Latest)
• OS: Linux (Ubuntu)
• Platform: Telegram

### Steps to Reproduce

Go to Telegram Channel Message

### Expected Behavior

It should render markdown in Telegram properly

### Actual Behavior

The Telegram bot integration currently fails to render Markdown formatting (like **bold**, *italic*, code blocks) correctly. Instead of formatted text, it displays the raw Markdown syntax (e.g., literal asterisks).

### Additional Context

_No response_

cu1ch3n:
## Summary

Telegram bot messages display raw Markdown syntax (literal asterisks, backticks) instead
of formatted text. This happens because `formatTextForPlatform()` lacks a Telegram-specific
path — it falls through to `escapeHtml()`, which strips HTML special characters but leaves
Markdown untouched. Since the plugin sends messages with `parse_mode: "HTML"`, the raw
Markdown is rendered as plain text.

**Changes:**

- **`ActionExecutor.ts`**: Route the `telegram` platform through `markdownToTelegramHtml()`
  in `formatTextForPlatform()`, consistent with how `lark` and `dingtalk` have their own
  conversion paths.
- **`TelegramAdapter.ts`**: Fix regex evaluation order in `markdownToTelegramHtml()` —
  fenced code blocks (`` ``` ``) must be matched before inline code (`` ` ``), otherwise
  the single-backtick pattern greedily consumes code block content.

Closes #1214

## Test plan

- [x] Trigger LLM responses containing **bold**, *italic*, `inline code`, fenced code
      blocks, and `[links](url)` via the Telegram bot
- [x] Confirm all formatting renders correctly in the Telegram client
- [x] Confirm fenced code blocks display as preformatted blocks, not broken inline code
- [x] `bun run test` passes with no regressions

### DIFF
--- a/src/channels/plugins/telegram/TelegramAdapter.ts
+++ b/src/channels/plugins/telegram/TelegramAdapter.ts
@@ -270,16 +270,22 @@ export function escapeMarkdownV2(text: string): string {
 
 /**
  * Convert Markdown to Telegram HTML
- * Basic conversion for common patterns
+ * Uses a placeholder approach to protect code content from bold/italic transformations
  */
 export function markdownToTelegramHtml(text: string): string {
   let result = escapeHtml(text);
 
-  // Code block must be processed before inline code to avoid partial matches
-  result = result.replace(/```(\w*)\n?([\s\S]*?)```/g, '<pre><code>$2</code></pre>');
+  // Extract and protect code blocks / inline code with placeholders
+  const placeholders: string[] = [];
+  const placeholder = (s: string) => {
+    const idx = placeholders.length;
+    placeholders.push(s);
+    return `\x00PH${idx}\x00`;
+  };
 
-  // Inline code
-  result = result.replace(/`([^`]+)`/g, '<code>$1</code>');
+  // Code blocks must be extracted before inline code
+  result = result.replace(/```(\w*)\n?([\s\S]*?)```/g, (_, _lang, code) => placeholder(`<pre><code>${code}</code></pre>`));
+  result = result.replace(/`([^`]+)`/g, (_, code) => placeholder(`<code>${code}</code>`));
 
   // Bold: **text** or __text__
   result = result.replace(/\*\*(.+?)\*\*/g, '<b>$1</b>');
@@ -291,6 +297,9 @@ export function markdownToTelegramHtml(text: string): string {
 
   // Links: [text](url)
   result = result.replace(/\[([^\]]+)\]\(([^)]+)\)/g, '<a href="$2">$1</a>');
+
+  // Restore placeholders
+  result = result.replace(/\x00PH(\d+)\x00/g, (_, idx) => placeholders[Number(idx)]);
 
   return result;
 }
--- a/tests/unit/channels/telegramMarkdown.test.ts
+++ b/tests/unit/channels/telegramMarkdown.test.ts
@@ -1,0 +1,99 @@
+/**
+ * @license
+ * Copyright 2025 AionUi (aionui.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, expect, it } from 'vitest';
+
+import { escapeHtml, markdownToTelegramHtml } from '@/channels/plugins/telegram/TelegramAdapter';
+
+describe('escapeHtml', () => {
+  it('should escape &, <, >', () => {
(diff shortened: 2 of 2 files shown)


goworm:
Nice fix on reordering the regex processing! However, there's still an issue: the content **inside** code blocks and inline code will still be incorrectly processed by the bold/italic regexes.

For example:

```
`file_name_test`
→ becomes <code>file_name_test</code>
→ then _(.+?)_ matches inside, producing <code>file<i>name</i>test</code>
```

The same applies to `*` inside code blocks — e.g. `glob_*_pattern` would get partial italic tags.

A common fix is to use a placeholder approach: extract code blocks and inline code first, replace them with placeholders, apply bold/italic transformations, then restore the original code content. Something like:

```typescript
export function markdownToTelegramHtml(text: string): string {
  let result = escapeHtml(text);

  // Extract and protect code blocks / inline code with placeholders
  const placeholders: string[] = [];
  const placeholder = (s: string) => {
    const idx = placeholders.length;
    placeholders.push(s);
    return `\x00PH${idx}\x00`;
  };

  result = result.replace(/```(\w*)\n?([\s\S]*?)```/g, (_, _lang, code) => placeholder(`<pre><code>${code}</code></pre>`));
  result = result.replace(/`([^`]+)`/g, (_, code) => placeholder(`<code>${code}</code>`));

  // Bold / Italic
  result = result.replace(/\*\*(.+?)\*\*/g, '<b>$1</b>');
  result = result.replace(/__(.+?)__/g, '<b>$1</b>');
  result = result.replace(/\*(.+?)\*/g, '<i>$1</i>');
  result = result.replace(/_(.+?)_/g, '<i>$1</i>');

  // Links
  result = result.replace(/\[([^\]]+)\]\(([^)]+)\)/g, '<a href="$2">$1</a>');

  // Restore placeholders
  result = result.replace(/\x00PH(\d+)\x00/g, (_, idx) => placeholders[Number(idx)]);

  return result;
}
```

This way, code content is fully protected from bold/italic transformations.

cu1ch3n:
Thanks for the review! I just fixed by using the placeholder approach you suggested. and I also added some unit tests covering the fix and basic formatting.

PR #1309, opened 2026-03-15
fix: show extension-contributed agents in conversation tab dropdown

iam-zekai:
## Summary

- Extension agents (those contributed via `aion-extension.json` with `backend="custom"` and `isExtension=true`) were not appearing in the `+` new-conversation dropdown inside `ConversationTabs`
- Fixes two related bugs in the same code path

## Root Cause

**Bug 1 — `splitConversationDropdownAgents` excluded all custom-backend agents:**
```ts
// Before
cliAgents: availableAgents.filter((agent) => agent.backend !== 'custom' && !agent.isPreset),

// After — extension agents (isExtension=true) are included
cliAgents: availableAgents.filter((agent) => (agent.backend !== 'custom' || agent.isExtension) && !agent.isPreset),
```
The filter was designed to exclude user-defined "custom agents" from the quick-add picker, but it also silently dropped extension-contributed agents which share the same `backend="custom"` value.

**Bug 2 — Duplicate menu keys when multiple extensions are installed:**
All extension agents generated the same key `cli:custom`, so clicking any of them would always resolve to the first extension agent in the list. Fixed by using a unique key format: `cli:ext:<extensionName>` for extension agents (regular agents continue using `cli:<backend>`).

## Changes

**`src/renderer/shared/agents/availableAgents.ts`**
- One-line filter change: `|| agent.isExtension` to include extension agents in the `cliAgents` bucket

**`src/renderer/pages/conversation/ConversationTabs.tsx`**
- Compute a unique `agentId` per agent: `ext:<extensionName>` for extensions, `<backend>` for built-ins
- Use `agentId` as the menu item key and for the lookup in `handleCreateConversation`
- Use `agent.extensionName ?? agent.backend` for logo lookup (forward-compatible with future logo mapping)

## Test Plan

- [ ] Install any extension that contributes an ACP adapter (e.g., Cursor Agent)
- [ ] Open a workspace conversation, click `+` in the tab bar
- [ ] Verify the extension agent appears under "CLI Agents"
- [ ] Click the extension agent → new conversation is created with the correct agent
- [ ] Verify existing built-in agents (Claude Code, Gemini CLI) still work correctly
- [ ] Install two extensions → verify each creates an independent conversation when clicked

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/renderer/pages/conversation/ConversationTabs.tsx
+++ b/src/renderer/pages/conversation/ConversationTabs.tsx
@@ -182,9 +182,13 @@ const ConversationTabs: React.FC = () => {
         let params;
 
         if (key.startsWith('cli:')) {
-          const backend = key.slice(4);
+          const agentId = key.slice(4);
+          // Extension agents use key format "ext:<extensionName>"; regular agents use their backend name.
           // [BUG-6] Null check: find() may return undefined
-          const agent = cliAgents.find((a) => a.backend === backend);
+          const agent = cliAgents.find((a) => {
+            const id = a.isExtension ? `ext:${a.extensionName ?? a.backend}` : a.backend;
+            return id === agentId;
+          });
           if (!agent) {
             Message.error(t('conversation.createFailed'));
             return;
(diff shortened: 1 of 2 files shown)


iam-zekai:
Withdrawing for now, will revisit after further design.

PR #1311, opened 2026-03-15
fix(acp): enhance CLI detection with fallback for macOS and improve agent reconciliation

lljr-laticia:
…gent reconciliation

- Added a fallback mechanism to resolve CLI paths for Codex on macOS when standard lookup fails.
- Refactored CLI availability checks into a dedicated method for better readability and maintainability.
- Implemented reconciliation of built-in agents to detect new CLI tools after initialization.
- Introduced unit tests to validate the new detection logic and fallback behavior.

# Pull Request

## Related Issues

- Closes #

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux
- [ ] My code follows the project's code style guidelines
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings or errors

## Additional Context

---

**Thank you for contributing to AionUi! 🎉**

### DIFF
--- a/src/agent/acp/AcpDetector.ts
+++ b/src/agent/acp/AcpDetector.ts
@@ -5,7 +5,10 @@
  */
 
 import { execSync } from 'child_process';
-import type { AcpBackendAll, PresetAgentType } from '@/types/acpTypes';
+import { existsSync } from 'fs';
+import { homedir } from 'os';
+import path from 'path';
+import type { AcpBackendAll, PotentialAcpCli, PresetAgentType } from '@/types/acpTypes';
 import { POTENTIAL_ACP_CLIS } from '@/types/acpTypes';
 import { ProcessConfig } from '@/process/initStorage';
 import { ExtensionRegistry } from '@/extensions';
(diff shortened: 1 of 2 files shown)


kaizhou-lab:
Hi, thanks for the contribution! The refactoring looks clean overall. Two things I'd like to see addressed before merging:

**1. Reconciliation should not run on every getter call.**

`reconcileBuiltinAgents()` is called inside `getDetectedAgents()` and `hasAgents()`, which are read-only methods invoked frequently by the UI (conversation list rendering, agent selector, etc.). Each call does synchronous `execSync` shell lookups with 1s timeouts — this will block the main process.

Instead, please provide an explicit `refresh()` method that callers can invoke when appropriate (e.g. after the user installs a new CLI, or on a manual "refresh agents" action). The getters should remain pure reads from the cached list.

**2. Fallback path design should be generic, not Codex-specific.**

Right now `resolveCliFallbackPath` hardcodes Codex.app paths only. If we add more agents in the future, we'd have to keep editing this function. A better approach would be to add an optional `fallbackPaths` field to the `PotentialAcpCli` type, so each CLI can declare its own fallback locations. This keeps the detector logic generic and the per-CLI knowledge in the data definition where it belongs.

PR #1321, opened 2026-03-16
fix(webserver): handle special characters in cookies properly

zhzy0077:
## Summary
- Use RFC6265 `cookie` package for WebSocket connection cookie parsing (bypasses Express middleware)
- Add try-catch for `decodeURIComponent` in client-side cookie parsing to handle malformed `%` sequences
- Fixes web requests hanging when cookies contain special characters like `%`

### Set cookie with special characters
```javascript
// In browser console
document.cookie = "test=!#$%&' *+-. ^_;path=/"
document.cookie = "aionui-session=token%with%percent;path=/"
```
### WebSocket request will hang/fail
```javascript
const ws = new WebSocket('ws://localhost:8080', [], {
  headers: { Cookie: 'test=!#$%&\' *+-. ^_;' }
});
ws.onopen = () => console.log('connected');
ws.onerror = (e) => console.log('error:', e);  // This fires
```
### Or via curl
```bash
curl -v -H "Cookie: test=!#\$%&' *+-. ^_;" http://localhost:8080/ws
# Request hangs or returns 500
```
## Expected vs Actual
| Cookie Value | Expected | Actual |
|-------------|----------|--------|
| `test=abc=def` | `{test: 'abc=def'}` | `{test: 'abc'}` |
| `test=test%XY` | `{test: 'test%XY'}` | `URIError: URI malformed` |
| `test=%` | `{test: '%'}` | `URIError: URI malformed` |

### DIFF
--- a/tests/unit/cookieParsing.test.ts
+++ b/tests/unit/cookieParsing.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import type { IncomingMessage } from 'http';
+
+// Mock AuthService before importing TokenMiddleware
+vi.mock('../../src/webserver/auth/service/AuthService', () => ({
+  AuthService: {
+    verifyToken: vi.fn(),
+    verifyWebSocketToken: vi.fn(),
+  },
+}));
+
+vi.mock('../../src/webserver/auth/repository/UserRepository', () => ({
+  UserRepository: { findById: vi.fn() },
+}));
+
+describe('extractWebSocketToken – cookie parsing with special characters', () => {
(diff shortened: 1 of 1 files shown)


kaizhou-lab:
Thanks for the fix! 🎉 Using the `cookie` package is much more robust than hand-rolling the parser.

I've pushed a commit with test cases covering special character scenarios (malformed `%`, `=` in values, etc.).

BTW — have you joined our community? We'd love to have you there:
- **Discord (English):** https://discord.gg/2QAwJn7Egx
- **微信群 (中文):** see [README](./resources/wechat_group_5.png)

PR #1322, opened 2026-03-16
feat(webui): expose WebUI settings to webui context

zhzy0077:
## Summary

- Exposes WebUI configuration panel in webui (browser) context
- Disables "Enable WebUI" toggle in webui since webui is always running
- Makes "Allow LAN Access" read-only in webui (managed by desktop app)
- **Key motivation**: Allows changing password on headless deployments where users access via webui URL directly

## UI Behavior

| Setting | Desktop | WebUI |
|---------|---------|-------|
| Enable WebUI | Functional | Disabled (checked) |
| Allow LAN Access | Functional | Read-only |
| Password Change | Functional | Functional |
| QR Code | Available | Available |
| Channels | Available | Available |


kaizhou-lab:
## Session Invalidation After Password/Username Change

Both `changePassword()` and `changeUsername()` call `AuthService.invalidateAllTokens()`, which rotates the JWT secret — **all existing sessions are immediately invalidated**, including the current user's own session.

### Impact in WebUI Context

Since this PR exposes the WebUI settings panel to the browser context, a user accessing via WebUI can now change their password or username. After doing so, their current session token becomes invalid immediately, and they get disconnected without any warning.

### Suggestion

Add a confirmation dialog before password/username change in WebUI context, e.g.:

> "Changing your password will invalidate all active sessions, including your current one. You will need to log in again with the new credentials. Continue?"

This way the user knows they'll be kicked out and can prepare (e.g., copy/remember the new password first).

kaizhou-lab:
A couple of suggestions:

1. `disabled` only guards the UI — users can still trigger `handleToggle` / `handleAllowRemoteChange` via DevTools. Consider adding `if (!isDesktop) return` at the top of those handlers as a safeguard.

2. The disabled switches have no explanation, which might confuse users. A tooltip like "This setting can only be changed in the desktop app" would go a long way.

kaizhou-lab:
Another concern: the desktop-side password change path (`WebuiService.changePassword`) does **not** require verifying the current password — this was intentional because the desktop app is the trusted host.

However, exposing this panel to the WebUI browser context means a remote user can now change the password **without knowing the current one**. This was the core reason we originally hid the WebUI settings panel in browser mode.

If we want to allow password change from WebUI, the user should be required to enter and verify their current password first (similar to the existing `/api/auth/change-password` route which already enforces this).

kaizhou-lab:
For headless deployments where the user forgot their password, we already provide a CLI reset command. See: https://github.com/iOfficeAI/AionUi/wiki/WebUI-Configuration-Guide

This might be a simpler and safer alternative to exposing password change in the WebUI panel.

PR #1351, opened 2026-03-17
fix(acp): clear session ID on stop to prevent stale session resume

### LINKED ISSUE #1303: [Bug]: 与cli问答时，点击停止按钮后，重发修正的问题，ai还是继续上个问题回复
### Platform

Windows

### AionUi Version

1.8.27

### Bug Description

发现ai回复方向有问题，想提前打断回复，但停止按钮似乎只是断开与cli的连接，重新发消息还会连带上个问题

### Steps to Reproduce

全自动模式下，点击停止按钮

### Expected Behavior

能够打断，断开与cli的连接应该单独有个按钮，并且只根据新问题形成答案，

### Actual Behavior

重新发消息还会连带上个问题

### Additional Context

_No response_

BillionClaw:
## Problem

When using CLI Q\u0026A (via ACP agents like Claude, Codex, etc.), clicking the stop button and then resending a corrected question causes the AI to continue the previous reply instead of starting fresh.

## Root Cause

When the stop button is clicked, the `stop()` method calls `disconnect()` which terminates the CLI process. However, the session ID (`acpSessionId`) stored in `this.extra` is NOT cleared.

When the user sends a new message:
1. `sendMessage()` detects the connection is lost and auto-reconnects via `start()`
2. `createOrResumeSession()` finds the stale `acpSessionId` and tries to resume the old session
3. The CLI resumes the previous conversation context, causing the AI to continue the old reply

## Fix

Clear `this.extra.acpSessionId` in the `stop()` method so subsequent messages create a new session instead of resuming the stale one.

## Changes

- `src/agent/acp/index.ts`: Added `this.extra.acpSessionId = undefined` in `stop()`

Fixes #1303

### DIFF
--- a/src/agent/acp/index.ts
+++ b/src/agent/acp/index.ts
@@ -483,6 +483,9 @@ export class AcpAgent {
     // Clear session-scoped caches when session ends
     this.approvalStore.clear();
     this.permissionRequestMeta.clear();
+    // Clear the session ID so the next message creates a fresh session
+    // instead of resuming the old one (fixes stop+resend bug)
+    this.extra.acpSessionId = undefined;
     // Emit finish event to reset frontend UI state
     this.onStreamEvent({
       type: 'finish',


kaizhou-lab:
Thanks for the fix! The analysis checks out — after `stop()`, the stale `acpSessionId` can cause `createOrResumeSession()` to resume the old session instead of creating a fresh one when the user resends.

One request: could you add a unit test that verifies `stop()` clears `acpSessionId`? Something like:

```ts
it('should clear acpSessionId after stop()', async () => {
  // setup agent with a known acpSessionId
  agent.extra.acpSessionId = 'old-session-id';
  await agent.stop();
  expect(agent.extra.acpSessionId).toBeUndefined();
});
```

This would prevent future regressions if `stop()` logic gets refactored.

PR #1511, opened 2026-03-19
feat: add block-no-verify hooks for Claude Code and Gemini CLI

### LINKED ISSUE #1509: Add block-no-verify to prevent AI agents from skipping git hooks
AionUi supports multiple AI coding agents (Claude Code, Gemini CLI, OpenCode, and more). These agents execute shell commands on behalf of users, including git operations. An agent that appends the no-verify flag to git commit or push will silently bypass all pre-commit and pre-push hooks — the checks that catch linting errors, test failures, and accidentally committed secrets.

This issue proposes integrating block-no-verify@1.1.2 (https://github.com/tupe12334/block-no-verify) into the AionUi agent configuration to block this class of hook bypass at the agent-execution layer.

**For Claude Code (.claude/settings.json):** A PreToolUse hook on the Bash matcher running 'npx --yes block-no-verify@1.1.2' will intercept any bash command before execution and reject those containing the disallowed flag.

**For Gemini CLI (.gemini/settings.json):** A BeforeTool hook on the run_shell_command matcher running the same command provides equivalent protection for Gemini-based sessions.

Since AionUi routes through many different underlying agents, adding this at the config level means protection is applied regardless of which agent a user is running.

Package: https://github.com/tupe12334/block-no-verify

_Disclosure: I am the author and maintainer of block-no-verify._

tupe12334:
Closes #1509. This PR adds agent-level enforcement to prevent AI coding agents from bypassing git hooks. It configures block-no-verify@1.1.2 (https://github.com/tupe12334/block-no-verify) as a hook in both Claude Code (.claude/settings.json with PreToolUse hook on Bash matcher) and Gemini CLI (.gemini/settings.json with BeforeTool hook on run_shell_command matcher). When any agent attempts to run a shell command containing the no-verify flag, the hook intercepts and blocks execution before it reaches git. _Disclosure: I am the author and maintainer of block-no-verify._

### DIFF
--- a/.gemini/settings.json
+++ b/.gemini/settings.json
@@ -1,0 +1,12 @@
+{
+  "hooks": {
+    "BeforeTool": [
+      {
+        "matcher": "run_shell_command",
+        "hooks": [
+          { "type": "command", "command": "npx --yes block-no-verify@1.1.2" }
+        ]
+      }
+    ]
+  }
+}


kaizhou-lab:
Thanks for the contribution! 🙏

We already have safeguards in place to prevent agents from skipping git hooks:

1. **Claude Code's built-in system prompt** explicitly instructs: *"Never skip hooks (--no-verify) unless the user has explicitly asked for it."*
2. **Our project's `CLAUDE.md`** reinforces this rule as well.

If you want an extra layer of protection without adding a third-party dependency, a simple inline hook in `.claude/settings.json` works:

```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": "if echo \"$TOOL_INPUT\" | grep -q -- '--no-verify'; then echo 'Blocked: --no-verify is not allowed'; exit 1; fi"
          }
        ]
      }
    ]
  }
}
```

This achieves the same goal with zero external dependencies.

PR #1745, opened 2026-03-26
fix: apply channel-configured model for gemini-cli ACP backend

JerryLiu369:
## Summary

- Channel model setting (e.g. `assistant.weixin.defaultModel`) was ignored for `gemini-cli` sessions — the subprocess always used its built-in default (`gemini-2.0-flash`)
- Root cause: `setModel` at session start was only called for the `claude` backend; no equivalent path existed for other ACP backends
- Fix threads the channel's configured model through `workerTaskManagerSingleton` → `AcpAgentManager.persistedModelId` → `AcpAgent.extra.currentModelId` → `connection.setModel()` at startup

## Changes

- `workerTaskManagerSingleton.ts`: pass `c.model?.useModel` as `currentModelId` when creating `AcpAgentManager` for `acp`-type conversations
- `AcpAgentManager.ts`: include `currentModelId: this.persistedModelId` in the `extra` passed to `new AcpAgent`
- `AcpAgent` (`index.ts`): add `else if (this.extra.currentModelId)` branch after the `claude` block — calls `setModel` at session start for any backend that has a configured model

## Test plan

- [ ] All existing unit tests pass (`npx vitest run tests/unit/channels/`)
- [ ] In WeChat channel with gemini-cli backend: set a non-default model in channel settings, start a new session, verify the correct model is used

### DIFF
--- a/src/process/agent/acp/index.ts
+++ b/src/process/agent/acp/index.ts
@@ -106,6 +106,8 @@ export interface AcpAgentConfig {
     acpSessionId?: string;
     /** Last update time of ACP session / ACP session 最后更新时间 */
     acpSessionUpdatedAt?: number;
+    /** Initial model ID to apply at session start (from channel config or persisted user choice) */
+    currentModelId?: string;
   };
   onStreamEvent: (data: IResponseMessage) => void;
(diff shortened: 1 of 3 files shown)


JerryLiu369:
Withdrawing for further investigation — model selection still not applying correctly in new WeChat sessions.

PR #1752, opened 2026-03-26
fix(weixin): channel mode fixes — gemini OAuth, model selection, tool UX

JerryLiu369:
## Summary

- Construct `google-auth-gemini` provider directly in server-side channel code (it is a frontend-only synthetic provider, never stored in model.config), so OAuth credentials are used instead of falling back to ADC/empty API key
- Allow reuse of existing `~/.gemini/oauth_creds.json` in channel mode when the user has already authenticated on desktop
- Apply channel-configured model for gemini-cli and other non-claude ACP backends at session start via `setModel()`
- Hide tool confirmation prompt in WeChat channel (yoloMode auto-approves tools; showing "Continue?" without interactive buttons is confusing)

## Root cause of WeChat ✕ Gemini hang

`google-auth-gemini` is built in `useModelProviderList.ts` on the frontend and never persisted to the database. The server-side `getChannelDefaultModel()` looked it up via `providerList.find()`, always got `undefined`, fell through to a Gemini provider with an empty API key, which triggered ADC (`Could not load default credentials`). `agent.send()` hung indefinitely → WeChat 5-minute timeout fired → no reply sent.

## Test plan

- [ ] Send a message from WeChat to the default Gemini agent — verify a reply is returned
- [ ] Confirm `~/.gemini/oauth_creds.json` is read (no ADC error in logs)
- [ ] Verify tool execution in WeChat shows progress text instead of a broken "Continue?" prompt
- [ ] Confirm gemini-cli uses the model configured in channel settings

### DIFF
--- a/src/process/channels/actions/SystemActions.ts
+++ b/src/process/channels/actions/SystemActions.ts
@@ -92,16 +92,25 @@ export async function getChannelDefaultModel(platform: PluginType): Promise<TPro
         }
 
         if (hasLocalCreds) {
-          // Use the Google Auth provider directly — local OAuth creds are available.
-          const googleAuthProvider = providerList.find((p) => p.id === GOOGLE_AUTH_PROVIDER_ID);
-          if (googleAuthProvider) {
-            return { ...googleAuthProvider, useModel: savedModel.useModel } as TProviderWithModel;
-          }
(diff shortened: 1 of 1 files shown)


JerryLiu369:
Withdrawing for further development and testing.

PR #1972, opened 2026-03-31
fix(codex): handle signal exits during startup

### LINKED ISSUE #1962: Codex startup exit handler misses signal-killed processes (ELECTRON-E4)
## Problem

In `CodexConnection.ts`, the exit handler during startup has condition `code !== 0 && code !== null` which fails when the process is killed by signal (code=null, signal=SIGTERM/SIGKILL). This causes:

1. `handleProcessExit` never runs — child reference dangles
2. The 5-second timeout fires with a generic unhelpful error: "Codex process failed to start or was killed during startup"

**Sentry issue:** ELECTRON-E4 — 9 events from multiple users (CA, CN, SG), last seen 5 hours ago.

## Root Cause

`null !== null` evaluates to `false`, so `code !== 0 && code !== null` is `false` when `code === null` (signal kill). The exit handler skips entirely.

## Fix

Remove the condition — any exit during the startup window is a failure. The error message now includes both code and signal for better diagnostics.

cdxiaodong:
## Summary

- treat any process exit during the Codex startup window as a startup failure, including signal-killed exits
- call the normal process-exit cleanup path for signal exits instead of leaving the child reference dangling
- add regression coverage for startup exits with SIGTERM

## Test plan

- [x] bun run lint
- [x] bun run format
- [x] bunx tsc --noEmit
- [x] bun run test

Fixes #1962

### DIFF
--- a/src/process/agent/codex/connection/CodexConnection.ts
+++ b/src/process/agent/codex/connection/CodexConnection.ts
@@ -215,10 +215,8 @@ export class CodexConnection {
         this.child.on('exit', (code, signal) => {
           // Reject the start promise immediately on unexpected exit during startup,
           // instead of waiting for the 5-second timeout with a generic error message.
-          if (code !== 0 && code !== null) {
-            reject(new Error(`Codex process exited during startup (code: ${code}, signal: ${signal || 'none'})`));
-            this.handleProcessExit(code, signal);
-          }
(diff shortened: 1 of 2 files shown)


piorpua:
## CI 检查未通过

以下 job 在本次自动化 review 时未通过，请修复：

| Job | 结论 |
|-----|------|
| Unit Tests (windows-2022) | ❌ FAILURE |

本次自动化 review 暂缓，待 CI 全部通过后将重新处理。

PR #2034, opened 2026-04-02
fix(conversation): move queued command execution gate to process

Ericwong5021:
## Summary

- move queued command execution gating from renderer to process to prevent the first queued command from being sent while a conversation page is still hydrating
- add a process-side execution service and IPC entry so renderer only requests execution instead of directly starting the next queued command
- update conversation platform send boxes to only apply local UI state after process accepts the queued command, and add regression coverage for busy/locked queue execution

## Root cause

When switching back to a conversation, history and queue state could restore before the runtime busy state was fully hydrated in renderer.

That allowed renderer-side auto-dequeue logic to incorrectly assume the conversation was idle and send the first queued command too early.

This change makes process the final authority for whether queued execution can start.

## Test plan

- [x] `bun run lint:fix`
- [x] `bun run format`
- [x] `bunx tsc --noEmit`
- [x] `bun run vitest tests/unit/conversationBridge.test.ts tests/unit/conversationBridge.tray.test.ts tests/unit/process/bridge/conversationBridge.sendMessage.test.ts tests/unit/renderer/conversationCommandQueue.dom.test.tsx tests/unit/conversationCommandExecutionService.test.ts`

## Notes

- `bun run test` still reports unrelated existing failures in this environment, including:
  - `tests/unit/configureChromium.test.ts`
  - `tests/unit/webuiQR.test.ts`
  - `tests/unit/webuiConfig.test.ts`

### DIFF
--- a/src/process/bridge/services/ActivitySnapshotBuilder.ts
+++ b/src/process/bridge/services/ActivitySnapshotBuilder.ts
@@ -118,7 +118,7 @@ export class ActivitySnapshotBuilder {
     for (const conversation of conversations) {
       const { backend, agentName } = resolveAgentIdentity(conversation);
       const task = this.taskManager.getTask(conversation.id);
-      const runtimeStatus = normalizeRuntimeStatus(task?.status || conversation.status);
+      const runtimeStatus = normalizeRuntimeStatus(task?.status);
       if (runtimeStatus === 'running' || runtimeStatus === 'pending') {
         runningConversations += 1;
       }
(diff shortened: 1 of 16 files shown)


piorpua:
## CI 检查未通过

以下 job 在本次自动化 review 时未通过，请修复：

| Job | 结论 |
|-----|------|
| Unit Tests (windows-2022) | ❌ FAILURE |

本次自动化 review 暂缓，待 CI 全部通过后将重新处理。

Ericwong5021:
分支落后主线，冲突太多，重新提交干净PR

PR #2349, opened 2026-04-13
feat(acp): managed install priority chain for agent discovery

tcp404:
## Summary

Managed install infrastructure for Hub-installed ACP agents: binary resolution, integrity verification, uninstall, and install source tracking.

## Key Changes

- **`ManagedInstallResolver`** — resolves agent binaries from Hub-managed directories (`bin/{cliCommand}` convention)
- **`AcpDetector` priority chain** — managed install → defaultCliPath → skip (no `which` needed for Hub agents)
- **`HubInstaller` rewrite** — remote-first download with SHA-256 content hash integrity verification
- **`HubStateManager` enhancements** — integrity-based update detection, `installSource` tracking (`extension` / `builtin`)
- **Agent uninstall** — full lifecycle: hook, managed dir cleanup, extension removal, state cleanup, registry reload
- **`AcpErrors` module** — structured error handling, fix ENOENT misdiagnosis (chdir vs binary missing)
- **`cacheKey` for custom agents** — `customAgentId` as cache key for model/config, instead of sharing one `'custom'` bucket
- **`bin/{cliCommand}` convention** — standardized binary layout replacing ad-hoc `installedBinaryPath`
- **Test coverage** — unit, integration, e2e, regression, fuzz tests for managed install and detection

## Status: Draft

All code changes are complete and type-checked. However, this PR is **blocked on auth flow support**:

Hub-installed agents live in managed directories — users cannot manually run login commands (e.g. `claude login`) since the binary isn't on their PATH. We need to implement an auth UI that:

1. Reads `authMethods` from the agent's `InitializeResponse`
2. Presents login options (OAuth/browser, terminal login, env var)
3. Triggers authentication via ACP `authenticate` RPC or terminal spawn
4. Retries session creation after successful auth

The ACP V2 rewrite (`src/process/acp/`) already has `AuthNegotiator`, `AUTH_REQUIRED` error handling, and `retryAuth()` infrastructure. The auth UI will be done in a follow-up branch.

Without auth flow, Hub-installed agents that require login will fail at first use. Agents using pre-configured env vars (e.g. `OPENAI_API_KEY`) work fine.

## Related Issues

- Closes #2348
- Closes #2373

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

PR #2374, opened 2026-04-13
fix(hub): enforce archive integrity verification

### LINKED ISSUE #2373: [Bug]: Hub installer skips archive integrity verification during install
## Bug Description

`HubInstaller.install()` resolves an extension archive and proceeds to extraction without verifying `dist.integrity`. As a result, a tampered bundled, cached, or remotely downloaded hub package can still be extracted and installed.

## Steps to Reproduce

1. Prepare a hub extension entry with a valid `dist.tarball` and `dist.integrity`
2. Modify the archive bytes so they no longer match the declared SHA-512 SRI
3. Trigger installation through the hub installer
4. Observe installation continuing to extraction instead of failing before install

## Root Cause

The hub schema requires `dist.integrity`, and `HubInstaller` already has a `verifyIntegrity()` helper, but the integrity check in the install flow was commented out. The helper also previously allowed unsupported algorithms to warn-and-skip instead of failing closed, which created a bypass path for untrusted runtime metadata.

## Affected Files

- `src/process/extensions/hub/HubInstaller.ts`
- `tests/unit/hubInstaller.test.ts`

## Diff

`+142 −11` across 2 files

amanharshx:
Closes #2373

## Summary

- Re-enable archive integrity verification in `HubInstaller.install()` before extraction
- Fail closed for missing, empty, malformed, unsupported, invalid-base64, or mismatched integrity metadata
- Add unit coverage for valid and invalid integrity scenarios, including verifying extraction is skipped on failure

## Motivation

The hub installer was accepting extension archives without validating the declared `dist.integrity` value. That allowed tampered bundled, cached, or remote packages to reach extraction and installation. Since hub metadata is runtime input, the install path must reject invalid integrity states deterministically instead of warning and continuing.

## Diff

`+142 −11` across 2 files

## Files changed

- `HubInstaller.ts` — re-enable integrity gate, normalize SRI with `trim()`, enforce strict SHA-512 validation, reject invalid base64 and mismatches
- `hubInstaller.test.ts` — add coverage for match, mismatch, missing integrity, whitespace-only integrity, unsupported algorithm, invalid base64, and no-extract-on-failure

## Testing

- `bun run test -- tests/unit/hubInstaller.test.ts` — passed
- `bunx tsc --noEmit` — passed
- `bun run format -- src/process/extensions/hub/HubInstaller.ts tests/unit/hubInstaller.test.ts` — passed
- `bun run lint:fix -- src/process/extensions/hub/HubInstaller.ts tests/unit/hubInstaller.test.ts` — clean except existing warning in `resolveZipPath()` for `await` inside loop

## Risks / Side effects

- Existing hub index entries with stale hashes will now fail installation immediately
- Existing non-`sha512-` integrity values will now fail instead of silently installing
- Strict validation may expose previously hidden metadata quality issues, but that is safer than allowing archive verification bypass

### DIFF
--- a/src/process/extensions/hub/HubInstaller.ts
+++ b/src/process/extensions/hub/HubInstaller.ts
@@ -118,8 +118,7 @@ export class HubInstallerImpl {
       const zipPath = await this.resolveZipPath(name, extInfo.dist.tarball, extInfo.bundled);
 
       // Step 2: Verify Integrity (SHA-512 SRI)
-      // TODO: 各平台校验有差异，先放在一边，后续完善
-      // await this.verifyIntegrity(zipPath, extInfo.dist.integrity);
+      await this.verifyIntegrity(zipPath, extInfo.dist.integrity);
 
       // Step 3: Extract (.zip)
       fs.mkdirSync(tempDir, { recursive: true });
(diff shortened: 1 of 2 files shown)


TCP404:
Hi @amanharshx, thank you for your contribution.

Regarding the verification issue, the code was originally commented out due to differences in `.zip` hash calculations across platforms. However, the Hub is still in a non-publicly launched state because we are currently refactoring the entire `ACP Discovery` feature. You can check #2349 for details. In that PR, the verification method has been changed to calculating the content's sha256, and it has already passed tests. If you are willing, you can participate in the work on #2349, or you can change it to the same processing method as #2349.

Thank you again for your contribution.

amanharshx:
@TCP404 @piorpua If this is already covered in #2349, feel free to take it from here and handle this PR as you prefer

PR #2378, opened 2026-04-14
fix(team): package team-mcp-stdio.mjs and add crash recovery

lixunqi12:
## Summary

- **Critical packaging fix**: `scripts/team-mcp-stdio.mjs` was referenced by `TeamMcpServer.getStdioConfig()` via `process.resourcesPath` but was never added to `electron-builder.yml`. In packaged installs, Claude CLI could not spawn the stdio bridge and team_* MCP tools silently never registered — team leader would report the MCP server as down. Adds the script to `extraResources`.
- **Crash recovery for team agents**: when an ACP agent process exits unexpectedly mid-turn, the team leader never received a terminal signal, leaving the turn stuck. `teamEventBus` now carries an `agentCrash` event that `TeammateManager` listens for; the leader is notified and the turn is finalized. Added consistent `role !== 'lead'` guard in the parse-failure branch.
- Switched `teamEventBus` import in `acp/index.ts` from lazy `require()` to static ES import so Vitest resolves the `@process/*` path alias correctly.

> Stacked on #2146 (the `team-mcp-stdio.mjs` file is introduced there). Merge after #2146.

## Test plan

- [x] `bunx tsc --noEmit` clean
- [x] `bunx vitest run tests/unit/acpTimeout.test.ts` — 20/20 pass
- [x] `bunx vitest run tests/unit/team` — 44/44 pass (11 skipped)
- [ ] Install the packaged build; start team mode; confirm leader sees team_* tools (no "MCP server is down" message).
- [ ] Kill an ACP subagent process mid-turn; confirm leader reports the crash and the turn finalizes instead of hanging.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -114,6 +114,11 @@ extraResources:
     to: bundled-aionrs
   - from: resources/hub
     to: hub
+  # Team MCP stdio bridge script — spawned by Claude CLI to expose team_* tools.
+  # Path resolved at runtime via process.resourcesPath + 'scripts/team-mcp-stdio.mjs'
+  # in TeamMcpServer.getStdioConfig(). Without this, Claude can't see team_* tools.
+  - from: scripts/team-mcp-stdio.mjs
+    to: scripts/team-mcp-stdio.mjs
 
 win:
   target:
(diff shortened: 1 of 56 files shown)


lixunqi12:
Closing as superseded. Upstream has since landed the main fixes this PR was targeting via a different route:
- #2401 refactored team wake logic (push-messages-only model)
- #2407 handled team-mcp-stdio packaging (renamed to team-guide-mcp-stdio)
- #2412 added crashed-worker detection

The three remaining independent UI/paste fixes have been split out into #2453 for clean review.

Thanks!

PR #2430, opened 2026-04-15
fix(conversation): restore image preview paths in chat history

lixunqi12:
## Summary

Images sent in chat (upload `+` button or paste) currently render as the "Image not found" placeholder because the path stored in the `[[AION_FILES]]` marker doesn't match the file on disk. This is the same class of bug that was fixed in #2361 and then lost when #2361 was reverted by #2370.

**Why #2370 happened (my reading):** The revert body only says *"Reverts #2361"*, no explanation. But the timeline shows #2361 also removed the user-facing `upload.saveToWorkspace` setting, and #2381 (day after the revert, same author as #2361) re-added that setting with a new default of `true`. The preview-path logic appears to have been collateral damage of a settings-UX disagreement, and was never re-landed.

This PR cherry-picks only the preview-path half of #2361 — no changes to the `saveToWorkspace` toggle or any other settings surface, so it shouldn't re-open whatever motivated #2370.

## Root cause

Two independent sources of mismatch on `main` today:

1. **Renderer strips the collision suffix.** `src/renderer/utils/file/messageFiles.ts` applied `AIONUI_TIMESTAMP_REGEX` to every path, turning `<workspace>/uploads/photo_aionui_1744…png` into `<workspace>/uploads/photo.png`. When there's an actual collision (paste the same screenshot twice, Gemini workspace already has that filename), `FilePreview.getImageBase64` ENOENTs and the fallback returns the placeholder SVG.
2. **Backend never rewrites the marker.** For non-Gemini agents `workspaceFiles` are cache-dir paths; for Gemini they may get a new `_aionui_<ts>` suffix during the workspace copy. Either way, the optimistic marker the renderer wrote at send time isn't corrected by the bridge, so the persisted chat bubble carries the wrong path.

The existing #1928 `FilePreview` retry loop (5 × 800ms) handles *timing* mismatches but can't rescue a *semantic* mismatch where the marker points at a path that will never exist — which is why the placeholder ends up sticking.

## Changes

- `src/renderer/utils/file/messageFiles.ts` — remove 3 `AIONUI_TIMESTAMP_REGEX` strips; pass external absolute paths through verbatim (previously rewritten to `${workspace}/${basename}`, which relied on a backend copy that no longer happens for non-Gemini agents).
- `src/renderer/components/media/FilePreview.tsx` — display `getCleanFileName(path)` (strips `_aionui_<ts>` for UI readability), load using the original `path`.
- `src/process/bridge/conversationBridge.ts` — reintroduce `buildCanonicalDisplayMessage` (from `7049d5fc`) and use it in `sendMessage.provider` so the persisted bubble + the agent-visible prompt both reference the actual paths the bridge is about to hand the agent.

No changes to: `upload.saveToWorkspace` setting, its i18n keys, the Gemini post-finish cleanup, the send-boxes under `src/renderer/pages/conversation/platforms/*`, or the retry logic from #1928.

## What should the reviewer pay attention to

1. **External-path passthrough behaviour change in `messageFiles.ts`.** If a user drags a file from *outside* the workspace, the marker is now the full absolute path instead of `${workspace}/${basename}`. For Gemini the bridge rewrites it with the post-copy workspace path, so that case is fine. For non-Gemini agents (ACP / Codex / NanoBot / OpenClaw / Remote) the agent now sees the real absolute path, which is what it actually needs to read the file — but please sanity-check whether any agent post-processes the prompt expecting workspace-relative paths.
2. **Optimistic-vs-canonical bubble.** The renderer's immediate `addOrUpdateMessage` in each sendbox still shows `buildDisplayMessage`'s optimistic output; the DB-persisted bubble (with the canonical marker from the bridge) replaces it when the task dispatches. In practice they match for the common cases; Gemini collision cases briefly show the pre-copy path before getting corrected. I considered propagating the canonical `displayMessage` back to each sendbox (as #2361 did via `return { success: true, data: { displayMessage } }`) but that touches 6 sendboxes and felt like scope creep — happy to expand if you'd rather kill the optimistic path outright.

## Test plan

- [x] `bun run format`
- [x] `bun run lint` (0 errors; only pre-existing `no-shadow` warnings in touched files)
- [x] `bunx tsc --noEmit`
- [x] `bunx vitest run tests/unit/messageFiles.test.ts tests/unit/FilePreview.dom.test.tsx tests/unit/process/bridge/conversationBridge.sendMessage.test.ts tests/unit/conversationBridge.test.ts tests/unit/geminiConversationBridge.test.ts tests/unit/acpConversationBridge.test.ts` — 40/40 pass, including new regression guards
- [x] Full `bunx vitest run` — 4115 pass / 7 fail, all 7 pre-existing on clean `origin/main` (Windows `fs.symlinkSync` EPERM + integration tests that require a built artifact)

## Manual smoke test (recommended before merge)

- Paste an image into an ACP or Gemini conversation with a workspace that already contains a file of the same name → image preview in the sent bubble should render (previously would show "Image not found" placeholder).
- Re-open the conversation → preview should still render from the persisted marker.
- Upload a file from outside the workspace via the `+` button for a non-Gemini agent → agent should receive a prompt containing the real absolute path.

### DIFF
--- a/src/process/bridge/conversationBridge.ts
+++ b/src/process/bridge/conversationBridge.ts
@@ -5,6 +5,7 @@
  */
 
 import { GeminiAgent, GeminiApprovalStore } from '@process/agent/gemini';
+import { AIONUI_FILES_MARKER } from '@/common/config/constants';
 import type { TChatConversation } from '@/common/config/storage';
 import type { IAgentManager } from '@process/task/IAgentManager';
 import type { IConversationService, CreateConversationParams } from '@process/services/IConversationService';
@@ -49,6 +50,27 @@ const VALID_CONVERSATION_TYPES = new Set<TChatConversation['type']>([
   'remote',
   'aionrs',
 ]);
+
+/**
+ * Replace the `[[AION_FILES]]` marker in a frontend-built display message with a canonical
+ * marker that references the actual file paths the backend will pass to the agent.
+ *
+ * The renderer's `buildDisplayMessage` writes an OPTIMISTIC marker before the bridge has a
+ * chance to copy/stage files. For Gemini we then copy uploads into the workspace and the
+ * resulting paths may differ from what the renderer guessed (e.g. a `_aionui_<ts>` collision
+ * suffix appears only on the workspace copy). Rewriting the marker here keeps the persisted
+ * chat bubble's preview path in sync with the real file on disk so FilePreview can resolve
+ * it on replay. Ported from PR #2361 (commit 7049d5fc) which was reverted in #2370 for
+ * reasons unrelated to the marker-rewrite logic (the revert targeted settings-toggle removal).
+ */
+const buildCanonicalDisplayMessage = (input: string, files: string[]): string => {
+  const markerIndex = input.indexOf(AIONUI_FILES_MARKER);
+  const text = markerIndex === -1 ? input : input.slice(0, markerIndex).trimEnd();
+  if (!files.length) {
+    return text;
+  }
+  return `${text}\n\n${AIONUI_FILES_MARKER}\n${files.join('\n')}`;
+};
 
 export function initConversationBridge(
   conversationService: IConversationService,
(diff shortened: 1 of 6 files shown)


piorpua:

PR #2452, opened 2026-04-16
fix(team): defer skipped wakes so mailbox messages aren't dropped mid-turn

lixunqi12:
## Summary

- Fixes an off-by-one user-visible bug in team conversations where a message sent to an agent that was busy processing a previous turn would sit unread in the mailbox until the *next* externally-triggered wake. Symptom: user sends msg1 → no reply → user sends msg2 → msg1's reply finally appears.
- Root cause: [`TeammateManager.wake()`](src/process/team/TeammateManager.ts) silently returned when `activeWakes` already held the slotId, dropping the wake request entirely. `finalizeTurn` had no mechanism to re-check pending work.
- Fix: record deferred wakes in a new `pendingWakes` Set and drain it at the end of `finalizeTurn`, so any mailbox message that arrived during the busy window is picked up in the same user-visible cycle.

## Before / After (timing trace)

**Before**
1. User sends msg1 → `wake(lead)` starts, `activeWakes.add(lead)`, reads mailbox, sends prompt
2. During the turn, some other signal (e.g. teammate `idle_notification`) triggers `wake(lead)` → deferred by `activeWakes`
3. *User sends msg2 mid-turn* → `wake(lead)` → `activeWakes.has(lead) === true` → **silently returns, msg2 stays unread**
4. `finalizeTurn` fires, sets idle, ends — msg2 still unread, nothing re-wakes
5. User sends msg3 → new wake reads **both msg2 + msg3** → agent replies to msg2 first → looks "offset by one"

**After**
1. Same as before through step 2
2. User sends msg2 mid-turn → `wake(lead)` → `activeWakes.has(lead)` → `pendingWakes.add(lead)` and return
3. `finalizeTurn` fires → after idle-notification logic, checks `pendingWakes.has(lead)` → **re-wakes → reads msg2 → agent replies**

## Changes

- `src/process/team/TeammateManager.ts`
  - New private `pendingWakes: Set<string>` field
  - `wake()`: skipped path now adds to `pendingWakes` instead of dropping silently (log changed from `SKIPPED` → `DEFERRED`)
  - `finalizeTurn()`: drain `pendingWakes` at the end, re-calling `wake()` when a deferred request is waiting
  - `handleAgentCrash()` / `removeAgent()` / `dispose()`: clear `pendingWakes` to prevent stale re-wakes on a dead slot
- `tests/integration/team-real-components.test.ts`
  - New regression test that writes a mailbox message during an active turn and verifies it's consumed by the drained wake after `finalizeTurn`. Verified this test fails on `main` and passes with the fix.

## Test plan

- [x] `bunx tsc --noEmit` clean
- [x] `oxlint` clean on changed files
- [x] `vitest run tests/integration/team-real-components.test.ts` — all 37 tests pass (36 existing + 1 new)
- [x] `vitest run tests/integration/team-stress-concurrency.test.ts` — all 21 tests pass
- [x] `vitest run tests/unit/process/team/` — all 9 tests pass
- [x] Regression test verified: fails on `main`, passes with fix applied

### DIFF
--- a/src/process/team/TeammateManager.ts
+++ b/src/process/team/TeammateManager.ts
@@ -39,6 +39,8 @@ export class TeammateManager extends EventEmitter {
 
   /** Tracks which slotIds currently have an in-progress wake to avoid loops */
   private readonly activeWakes = new Set<string>();
+  /** Slots whose wake() was deferred because activeWakes was set; drained after finalizeTurn */
+  private readonly pendingWakes = new Set<string>();
   /** Timeout handles for active wakes, keyed by slotId */
   private readonly wakeTimeouts = new Map<string, ReturnType<typeof setTimeout>>();
   /** O(1) lookup set of conversationIds owned by this team, for fast IPC event filtering */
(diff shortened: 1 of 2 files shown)


piorpua:

PR #2426, opened 2026-04-15
fix(team): prevent 300s LLM timeout when teammates stand by

lixunqi12:
## Summary

Fixes a real-world failure mode where a teammate (most often `codex`) was assigned a task with a "stand by until X finishes" instruction, kept its LLM stream open while waiting, and got killed by the provider's hard ~300s request timeout — surfaced to the user as `LLM request timed out after 300 seconds` and the agent marked failed.

Two prompt-level changes, no runtime code changes:

- **`teammatePrompt.ts` — new "Standing By" section.** Explicitly instructs the agent that "waiting" / "standing by" means ending the current turn, not generating filler text in a live stream. Acknowledge once via `team_send_message`, then stop generating. The mailbox + wake mechanism re-activates the agent the moment work arrives.
- **`leadPrompt.ts` — new "Sequencing Dependent Work" section.** Tells the lead to dispatch dependent tasks sequentially as prerequisites complete, instead of dispatching them in parallel with "wait for X" instructions, so downstream agents are never asked to keep an LLM stream open while blocked on another teammate.

Pairs naturally with iOfficeAI/AionUi#2425 (the watchdog now reports a stuck agent to the lead instead of silently flipping it back to idle), but the two PRs are independent — either can land without the other.

## Test plan

- [x] `bun run lint` — 0 errors.
- [x] `bunx tsc --noEmit` — clean.
- [x] `bunx vitest run tests/unit/team-TeammateManager.test.ts tests/integration/team-stress-concurrency.test.ts` — 81 / 81 pass (no behavior change to the runtime, prompts only).
- [ ] Manual: dispatch a "code review after implementer" workflow with a codex teammate; verify the reviewer ends its turn after acknowledging instead of holding an LLM stream until the 300s server-side timeout.

PR #2505, opened 2026-04-17
fix(team): ensure teammate output reaches lead without wake loop

lixunqi12:
## Summary

Three chained fixes so non-lead teammates can reliably communicate with the lead across a turn:

1. **Deferred wakes (`e4aab55`)** — mailbox messages that arrive while a teammate is mid-turn are no longer dropped. `wake()` now records the request in `pendingWakes` and `finalizeTurn` drains it so the next message lands in the same user-visible cycle (not the one after).

2. **Auto-captured fallback (`399bef2`)** — when a non-lead teammate finishes a turn without calling `team_send_message`, the lead used to receive only the literal string `Turn completed` and was blind to anything the teammate actually produced. The manager now keeps a tail-biased response buffer (`MAX_RESPONSE_CAPTURE_CHARS = 3000`), snapshots it at `finalizeTurn`, and forwards it to the lead prefixed with `[auto-captured fallback …]` so the lead can distinguish it from an explicit report. `leadPrompt.ts` gains a section explaining both flavors. An explicit-send flag (set by `TeamMcpServer.handleSendMessage` whenever a teammate writes to the lead) suppresses the fallback that turn to avoid duplicate reporting.

3. **Empty-fallback wake loop (`31c1059`)** — a teammate whose turn produced no text and no explicit report now writes the empty-fallback notification but does NOT trigger `maybeWakeLeaderWhenAllIdle`. Previously the empty notification would still wake the lead, which would re-dispatch the same task, producing another empty turn, and so on. When every teammate is only producing empty fallbacks we now prefer the team to stall (the empty notifications remain in the mailbox as history) rather than spin.

## Why these belong together

The three commits are the same arc: \#1 keeps messages from being lost, \#2 surfaces output the teammate didn't explicitly report, \#3 makes sure \#2's safety-net doesn't itself produce an infinite loop. Reviewing or reverting them independently would leave the system in a partially-consistent state.

## Interaction with recent upstream changes

- \#2425 (silent-agent watchdog) — orthogonal. The watchdog fires when a teammate streams nothing for 60s; our changes only touch `finalizeTurn` (called from `type: 'finish' | 'error'`) and the `handleResponseStream` content-capture branch (runs before the watchdog heartbeat). Merge resolution keeps the watchdog's `resetWakeTimeout` + `handleInactivityTimeout` intact.
- \#2426 (stand-by prompt) — complementary. That fix prevents teammates from sitting in open LLM streams; this PR addresses what happens after the turn ends.
- `safeWake()` helper from \#2425 is preserved and used for all four teammate-send code paths in `TeamMcpServer.handleSendMessage`.

## Test plan

- [x] `bunx vitest run tests/unit/team-TeammateManager.test.ts` — 85 passed (covers capture buffer cap/tail-bias, marker prefix, explicit-send dedup, empty-fallback no-wake, pending-wakes drain, cleanup paths, 200-turn stress test)
- [x] `bunx vitest run tests/unit/team-TeamMcpServer.test.ts tests/integration/team-real-components.test.ts tests/integration/team-stress-concurrency.test.ts` — 96 passed (covers explicit-send notification on direct/broadcast/shutdown paths; full-chain turn completion with content)
- [x] Manual dev-mode run with a real multi-agent team — previously looped on empty fallbacks, now stalls cleanly and resumes on the next user input

### DIFF
--- a/src/process/team/TeamSession.ts
+++ b/src/process/team/TeamSession.ts
@@ -64,6 +64,10 @@ export class TeamSession extends EventEmitter {
         this.teammateManager.removeAgent(slotId);
       },
       wakeAgent: (slotId: string) => this.teammateManager.wake(slotId),
+      // Tell the manager when a teammate explicitly delivers to the leader
+      // so it can suppress the fallback idle_notification this turn and
+      // avoid duplicating the teammate's report in leader's mailbox.
+      notifyExplicitSendToLead: (fromSlotId: string) => this.teammateManager.markExplicitSendToLead(fromSlotId),
     });
   }
 
--- a/src/process/team/TeammateManager.ts
+++ b/src/process/team/TeammateManager.ts
@@ -49,9 +49,44 @@ export class TeammateManager extends EventEmitter {
   private readonly finalizedTurns = new Set<string>();
   /** Maps slotId → original name before rename, for "formerly: X" hints in prompts */
   private readonly renamedAgents = new Map<string, string>();
+  /**
+   * Accumulates the assistant text a non-lead teammate streams during a turn,
+   * keyed by conversationId. Used as a SAFETY NET: when a teammate finishes
+   * their turn WITHOUT explicitly calling team_send_message to the leader,
+   * we forward the captured text via idle_notification so the leader isn't
+   * blind. If the teammate DID call team_send_message to the leader, the
+   * explicit (curated) message is authoritative and we skip the fallback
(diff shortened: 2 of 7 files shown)


kaizhou-lab:
## CI 检查未通过

以下 job 在本次自动化 review 时未通过,请修复:

| Job | 结论 |
|-----|------|
| Code Quality | ❌ FAILURE |

本次自动化 review 暂缓,待 CI 全部通过后将重新处理。

piorpua:

PR #2528, opened 2026-04-17
fix(auth): rotate refreshed session tokens

### LINKED ISSUE #2527: [Bug]: auth refresh endpoint does not rotate cookie-backed sessions
## Bug Description

`POST /api/auth/refresh` returns a new JWT in the response body, but does not rewrite the `aionui-session` cookie. Cookie-authenticated WebUI clients stay on the stale token. The refresh service also claims to skip expiry checks, but currently rejects expired tokens.

## Steps to Reproduce

1. Log in to WebUI with a cookie-backed session
2. Call `POST /api/auth/refresh`
3. Observe the response contains a new token
4. Make another authenticated request from the browser
5. Observe the browser still sends the old session cookie

## Root Cause

The refresh route only reads `req.body.token` and never rotates the session cookie on success. In addition, `refreshToken()` used normal token verification semantics, so expired-but-otherwise-valid tokens could not be refreshed. After adding strict rotation, a unique token claim was also needed so same-second refreshes do not mint an identical JWT and immediately blacklist it.

## Affected Files

- `src/process/webserver/routes/authRoutes.ts`
- `src/process/webserver/auth/service/AuthService.ts`
- `tests/unit/webserver/authRoutesRefresh.test.ts`
- `tests/unit/webserver/authServiceRefresh.test.ts`

## Diff

`+158 −14` across 4 files

amanharshx:
Closes #2527

## Summary

- Read refresh token from `req.body.token` or request auth state
- Rewrite the session cookie after a successful refresh
- Allow refresh for expired but otherwise valid session JWTs
- Rotate the old token after issuing a replacement token
- Add a unique `sessionId` claim so refreshed JWTs cannot collide with the old token
- Add regression coverage for cookie fallback, cookie rotation, expired-token refresh, and token blacklisting

## Motivation

Cookie-authenticated clients could not actually use the refresh endpoint, and the refresh service behavior did not match its documented contract. This change makes refresh work for WebUI sessions and aligns the implementation with the intended token renewal flow.

## Diff

`+158 −14` across 4 files

## Testing

- `bun run vitest tests/unit/webserver/authRoutesRefresh.test.ts tests/unit/webserver/authServiceRefresh.test.ts`
- `bun run format src/process/webserver/routes/authRoutes.ts src/process/webserver/auth/service/AuthService.ts tests/unit/webserver/authRoutesRefresh.test.ts tests/unit/webserver/authServiceRefresh.test.ts`

## Risks / Side effects

- Successful refresh now invalidates the previous token
- JWT payloads now include a unique `sessionId` claim
- `bunx tsc --noEmit` still fails on unrelated pre-existing ACP / WeCom dependency issues

### DIFF
--- a/src/process/webserver/auth/service/AuthService.ts
+++ b/src/process/webserver/auth/service/AuthService.ts
@@ -12,6 +12,7 @@ import { UserRepository } from '../repository/UserRepository';
 import { AUTH_CONFIG } from '../../config/constants';
 
 interface TokenPayload {
+  sessionId?: string;
   userId: string;
   username: string;
   iat?: number;
@@ -234,6 +235,7 @@ export class AuthService {
    */
   public static async generateToken(user: Pick<AuthUser, 'id' | 'username'>): Promise<string> {
     const payload: TokenPayload = {
+      sessionId: this.generateSessionId(),
       userId: user.id,
       username: user.username,
     };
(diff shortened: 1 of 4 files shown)


amanharshx:
@piorpua Closing this. Would've been nice if existing PRs had been checked before merging a duplicate.

PR #2587, opened 2026-04-22
fix(shell): add ~/.local/bin to Windows extra tool paths

### LINKED ISSUE #2538: Windows: AionUI fails to launch Claude Code when installed via new native installer (~/.local/bin/claude.exe)
### Describe the bug

AionUI cannot launch Claude Code on Windows, even though Claude Code is already installed locally and works correctly in terminal.

The error shown in AionUI is:

```text
'claude' CLI not found. Please install it or update the CLI path in Settings.
ENOENT: Bun could not find a file, and the code that produces this error is missing a better error.
```

### To Reproduce

1. Install Claude Code on Windows (using the new native installer, which places `claude.exe` under `%USERPROFILE%\.local\bin\`)
2. Open AionUI
3. Try to use Claude Code from within AionUI
4. AionUI fails with the error above

### Expected behavior

AionUI should detect and launch Claude Code successfully when `claude` is already installed and available from the terminal / PATH, regardless of whether it was installed via the old npm global method or the new native installer.

### Actual behavior

AionUI reports that the `claude` CLI is not found and throws `ENOENT` from the bundled Bun runtime, even though `claude` works correctly outside AionUI.

### Environment

- OS: Windows 10
- AionUI version: `v1.9.17`
- Claude Code version: `2.1.112`

### Verification

In PowerShell:

```powershell
where.exe claude
```

Output:

```text
C:\Users\[User Name]\AppData\Roaming\npm\claude
C:\Users\[User Name]\AppData\Roaming\npm\claude.cmd
C:\Users\[User Name]\.local\bin\claude
C:\Users\[User Name]\.local\bin\claude.exe
```

Also:

```powershell
claude --version
```

Output:

```text
2.1.112 (Claude Code)
```

### Possible cause

Claude Code's official Windows installer has moved the CLI location between versions:

| Install method | Executable path | Extension |
| --- | --- | --- |
| Old (npm global) | `%APPDATA%\npm\claude.cmd` | `.cmd` (shell wrapper) |
| New (native installer) | `%USERPROFILE%\.local\bin\claude.exe` | `.exe` (native binary) |

On this machine **both** paths exist side by side, which confirms that `where claude` / detection alone is not enough — AionUI must also be able to actually *spawn* the binary from the bundled Bun runtime.

Looking at the AionUI source I can see a few suspicious spots that may be related:

- `src/common/types/acpTypes.ts` (around line 342): the Claude backend is registered with `cliCommand: 'claude'` only — no `.cmd` / `.exe` disambiguation, which is fine for detection but may cause problems when the bundled Bun process later tries to resolve the executable.
- `src/process/utils/shellEnv.ts` (Windows `candidates` list, around lines 311–374): `%APPDATA%\npm` is listed first as `// npm global packages (most common - installed with Node.js)`, and `~/.local/bin` is included but labeled only `// local bin (pip, pipx, etc.)`. The comment suggests `.local\bin` was added for Python tooling, not for Claude Code's native installer, so its ordering/priority may not have been reviewed with Claude Code in mind.
- `src/process/agent/acp/acpConnectors.ts` `prepareClaude` (around line 422) launches Claude via `bunx @zed-industries/claude-agent-acp`, which in turn needs to locate `claude` itself. The user-visible error `ENOENT: Bun could not find a file` comes from this bundled-bun spawn path, which suggests bundled Bun's child process is unable to resolve `claude` even though `where claude` in the user's terminal succeeds.
- `src/process/agent/acp/AcpConnection.ts` (around line 300): the generic ENOENT → `'<cli>' CLI not found` error message is emitted without distinguishing between "CLI is truly missing" and "CLI exists but bundled Bun can't spawn it on Windows", which makes this class of issue hard to diagnose.

So the combination of symptoms (detection works in terminal, but AionUI fails with a Bun ENOENT) points to a Windows CLI resolution / spawn issue inside the bundled Bun path, specifically for the newer `.local\bin\claude.exe` layout.

### Suggested fix

Please review Claude Code detection/launch on Windows to support the new installer layout, specifically:

- Explicitly document and support `%USERPROFILE%\.local\bin\claude.exe` as a first-class Claude Code install location (not only as a pip/pipx path).
- Ensure the bundled Bun runtime inherits a PATH that includes `%USERPROFILE%\.local\bin` when spawning `@zed-industries/claude-agent-acp`, and that it can resolve both `.cmd` and `.exe` variants (respect `PATHEXT`).
- Improve the error in `AcpConnection.ts` so that "CLI not found" vs. "CLI found but bundled Bun could not spawn it" are distinguishable — currently both surface as the same message, which hides the real cause.
- Consider adding an explicit "detected Claude Code path" line in Settings / diagnostics so users can see which `claude` binary AionUI is actually trying to launch.

### Additional context

This does not appear to be a Claude Code installation problem, because `claude --version` works correctly in PowerShell and both the old npm path and the new `.local\bin` path are present on the machine.

The issue seems to be that AionUI (or the bundled Bun runtime it uses to launch `@zed-industries/claude-agent-acp`) cannot find or launch the executable even though it exists and is accessible from PATH — most likely because Claude Code's recent switch from `%APPDATA%\npm\claude.cmd` to `%USERPROFILE%\.local\bin\claude.exe` was not accounted for in the Windows launch path.

octo-patch:
Fixes #2538

## Problem

Claude Code's native Windows installer (≥ 2.1.112) places `claude.exe` at `%USERPROFILE%\.local\bin\` instead of the legacy `%APPDATA%\npm\` path. Since `~/.local/bin` was not included in `getWindowsExtraToolPaths()`, AionUi could not detect or launch Claude Code when installed via the new native installer.

The POSIX equivalent (`~/.local/bin`) was already listed in `getPosixExtraToolPaths()`, but the Windows function was missing it.

## Solution

Add `path.join(homeDir, '.local', 'bin')` to the Windows extra tool path candidates in `getWindowsExtraToolPaths()` (shellEnv.ts). This ensures both the legacy npm-global path (`%APPDATA%\npm\`) and the new XDG-style user-local install location (`%USERPROFILE%\.local\bin\`) are included in AionUi's enhanced PATH.

## Testing

- Added a unit test verifying that `~/.local/bin` is appended to the enhanced PATH when the directory exists on Windows.
- All 31 existing `shellEnv` unit tests still pass.
- `bunx tsc --noEmit` passes with no errors.
- `bun run lint` passes with 0 errors.

### DIFF
--- a/src/process/utils/shellEnv.ts
+++ b/src/process/utils/shellEnv.ts
@@ -344,6 +344,10 @@ function getWindowsExtraToolPaths(): string[] {
     'C:\\cygwin\\bin',
     // bun global packages
     getBunGlobalBinDir(),
+    // Claude Code native installer and other XDG-style user-local tools.
+    // The new Claude Code Windows installer (≥ 2.1.112) places claude.exe at
+    // %USERPROFILE%\.local\bin rather than the old %APPDATA%\npm location.
+    path.join(homeDir, '.local', 'bin'),
   ];
 
   return candidates.filter((p) => existsSync(p) && !currentPath.includes(p));
--- a/tests/unit/shellEnv.test.ts
+++ b/tests/unit/shellEnv.test.ts
@@ -15,6 +15,7 @@
  */
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import os from 'os';
 import path from 'path';
 
 const mocks = vi.hoisted(() => ({
@@ -371,6 +372,38 @@ describe('getEnhancedEnv Windows extra paths (cross-platform mock)', () => {
     // Should appear exactly once (from process.env.PATH), not duplicated
     const occurrences = result.PATH.split(';').filter((p) => p === GIT_USR_BIN).length;
     expect(occurrences).toBe(1);
+  });
+
+  it('appends ~/.local/bin when it exists (Claude Code native installer path)', async () => {
+    Object.defineProperty(process, 'platform', { value: 'win32' });
+    process.env.PATH = 'C:\\Windows\\System32';
+    process.env.APPDATA = 'C:\\Users\\test\\AppData\\Roaming';
+    process.env.LOCALAPPDATA = 'C:\\Users\\test\\AppData\\Local';
+    process.env.ProgramFiles = 'C:\\Program Files';
+
+    const LOCAL_BIN = path.join(os.homedir(), '.local', 'bin');
+
+    vi.doMock('fs', async () => {
+      const actual = await vi.importActual<typeof import('fs')>('fs');
+      return {
+        ...actual,
+        existsSync: vi.fn((p: string) => p === LOCAL_BIN),
+        readdirSync: actual.readdirSync,
+        accessSync: actual.accessSync,
+      };
+    });
+
+    vi.doMock('child_process', () => ({
+      execFileSync: vi.fn().mockImplementation(() => {
+        throw new Error('skip shell');
+      }),
+      execFile: vi.fn(),
+    }));
+
+    const { getEnhancedEnv } = await import('@process/utils/shellEnv');
+    const result = getEnhancedEnv();
+
+    expect(result.PATH).toContain(LOCAL_BIN);
   });
 
   it('skips shell env loading on Windows and relies on extra tool paths', async () => {


piorpua:

octo-patch:
Thanks for the heads-up about the repo split. Closing this PR — happy to re-open against AionUi/AionCore if the underlying issue is still relevant after the restructure.

PR #2588, opened 2026-04-22
fix(acp): use where.exe in sync Windows agent availability check

### LINKED ISSUE #2523: [Bug]: Local agents not detected in Windows
### Platform

Windows

### AionUi Version

1.9.17

### Bug Description

In Windows 10, AionUI usually fails to detect codex or claude code. After some research, I guess it might come from the legacy support of `where` command.

AionUI uses `where` in Windows powershell to detect local agents. However, in Windows powershell, `where` is an alias to Where-Object for object filtering. 

To find the location of a command same as which in MacOS, we could use where in **cmd** rather than in powershell. Luckily, we could still use where.exe(located in C:\Windows\System32\where.exe) or Get-Command in Windows powershell to get the result.

Windows中AionUI采用powershell的where来查找cli工具如codex、claude code。cmd中的where确实能查找可执行文件的路径，但powershell的where是Where-Object的别名，无法查找可执行文件。powershell中应该用where.exe(C:\Windows\System32\where.exe)或者Get-Command查找可执行文件的路径。

Image

### Steps to Reproduce

Open up AionUI, and it failed to show built-in codex or claude code. Only show Aion CLI and Gemini Cli.

### Expected Behavior

AionUI should autodetect codex or claude code, which is already installed.

### Actual Behavior

AionUI fails to detect Codex or Claude Code.
AionUI通常无法检测到Codex或者Claude Code。

### Additional Context

_No response_

octo-patch:
Fixes #2523

## Problem

The synchronous CLI availability check (`batchCheckCliAvailabilitySync`) runs via `execSync()`, which invokes the shell specified by the `COMSPEC` environment variable. When `COMSPEC` is set to PowerShell (a non-standard but possible configuration), the bare `where` command resolves to PowerShell's `Where-Object` alias rather than the Windows system binary, causing CLI lookups to silently fail.

This means agents like `openclaw` and `nanobot` may not be detected on Windows when the shell is PowerShell.

## Solution

Replace `'where'` with `'where.exe'` in the sync Windows detection path so the correct system binary (`C:\Windows\System32\where.exe`) is invoked explicitly, regardless of which shell `COMSPEC` points to.

The async detection path (`batchCheckCliAvailability`) already uses `safeExecFile('where', ...)` which bypasses the shell entirely and is unaffected by this issue.

## Testing

- `bunx tsc --noEmit` passes with no errors.
- `bun run lint` passes with 0 errors.
- `bun run format` passes cleanly.

### DIFF
--- a/src/process/agent/acp/AcpDetector.ts
+++ b/src/process/agent/acp/AcpDetector.ts
@@ -122,7 +122,12 @@ class AcpDetector {
 
     const { execSync } = require('child_process') as typeof import('child_process');
     const isWindows = process.platform === 'win32';
-    const whichCommand = isWindows ? 'where' : 'which';
+    // Use where.exe explicitly on Windows so the lookup works regardless of
+    // whether the current shell is cmd.exe or PowerShell. In PowerShell,
+    // `where` is an alias for Where-Object (object filtering) and will not
+    // locate executables. `where.exe` always invokes the correct Windows
+    // system binary (C:\Windows\System32\where.exe).
+    const whichCommand = isWindows ? 'where.exe' : 'which';
     const found = new Set<string>();
 
     for (const cmd of safe) {


piorpua:

octo-patch:
Thanks for the heads-up about the repo split. Closing this PR — happy to re-open against AionUi/AionCore if the underlying issue is still relevant after the restructure.

PR #2589, opened 2026-04-22
fix(media): retry transient "image not found" placeholder in chat messages

lixunqi12:
## Summary

Chat-attached images intermittently rendered as the gray "Image not found" SVG instead of the real image. The root cause is a race between chat-message rendering and the backend finishing the write of a pasted/uploaded image to disk:

- `fsBridge.getImageBase64` catches `fs.readFile` errors (typically `ENOENT`) and **resolves** with a placeholder data-URL SVG rather than throwing.
- `LocalImageView` only reacted to a rejected promise, so the placeholder was committed to `url` and rendered permanently.
- `FilePreview` (the compose-bar preview) already worked around this by substring-matching the base64 placeholder and retrying, but `LocalImageView` (the in-conversation view) did not.

This PR:

1. Extracts the placeholder marker + retry constants into a shared `src/renderer/components/media/imagePlaceholder.ts` module (removes the duplicate in `FilePreview`).
2. Adds the same retry policy to `LocalImageView` — up to 5 attempts, 800 ms apart, with `cancelled` flag + `clearTimeout` cleanup on unmount and on `absolutePath` change.
3. Adds `tests/unit/LocalImageView.dom.test.tsx` mirroring the existing `FilePreview` coverage (happy path, retry-then-succeed, max-retries exhausted, unmount cleanup, promise rejection).

No behavior change when the image is available on first read; the retry only kicks in when the placeholder marker is detected.

## Test plan

- [x] `bun run lint:fix` — 0 errors
- [x] `bun run format` / `bun run format:check` — clean
- [x] `bunx vitest run tests/unit/LocalImageView.dom.test.tsx tests/unit/FilePreview.dom.test.tsx` — **10 / 10 passing**
- [x] `bun run i18n:types` + `node scripts/check-i18n.js` — pass
- [ ] Manual smoke: paste a screenshot into a Claude Code conversation and confirm it renders (previously would show "Image not found")

### DIFF
--- a/src/renderer/components/media/FilePreview.tsx
+++ b/src/renderer/components/media/FilePreview.tsx
@@ -10,14 +10,9 @@ import { getFileExtension } from '@/renderer/services/FileService';
 import { ipcBridge } from '@/common';
 import { Image } from '@arco-design/web-react';
 import fileIcon from '@/renderer/assets/icons/file-icon.svg';
+import { IMAGE_RETRY_DELAY_MS, MAX_IMAGE_RETRIES, isImageNotFoundPlaceholder } from './imagePlaceholder';
 
 const IMAGE_EXTS = new Set(['.jpg', '.jpeg', '.png', '.gif', '.bmp', '.webp', '.svg']);
-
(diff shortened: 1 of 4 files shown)


piorpua:

PR #2594, opened 2026-04-22
fix(acp): defer prompt timeout while tool calls are in flight

lixunqi12:
## Summary

- Long-running tool calls (multi-minute web fetches, large file reads) send `tool_call` at start and `tool_call_update` at completion with no heartbeat in between. The 5-minute prompt timer would fire mid-tool and cancel the turn even though the agent was still productively working.
- `PromptExecutor` now tracks in-flight `toolCallId`s. When the timer fires while the pending set is non-empty it restarts itself instead of calling `cancel()`, so the timeout only fires on true agent silence.
- `AcpSession.handleMessage` forwards `tool_call` / `tool_call_update` updates into `trackToolCallLifecycle` so the executor sees both start and completion/failure events.

## Test plan

- [x] 13 new unit tests in `PromptExecutor.test.ts` cover: tracking start, tracking completion, tracking failure, ignoring updates without `toolCallId`, timer deferral when tools pending, timer firing after all tools complete, and idempotent add/remove.
- [x] `bunx vitest run tests/unit/process/acp/session/PromptExecutor.test.ts` — all 13 pass.
- [x] `bunx tsc --noEmit` clean.
- [x] `bun run lint` clean on changed files.
- [ ] Manual smoke: run a Claude Code session that invokes a long (>5min) tool, confirm no "Prompt timed out" signal fires mid-tool.

Related: complements #1702 (process-liveness keepalive). The keepalive prevents timeouts when the agent process is alive but idle; this PR prevents timeouts when the agent is actively waiting on a tool.

### DIFF
--- a/src/process/acp/session/AcpSession.ts
+++ b/src/process/acp/session/AcpSession.ts
@@ -353,6 +353,7 @@ export class AcpSession {
       }
     }
 
+    this.promptExecutor.trackToolCallLifecycle(update);
     this.promptExecutor.resetTimer();
     const messages = this.messageTranslator.translate(notification);
     for (const msg of messages) {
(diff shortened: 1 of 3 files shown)


piorpua:

PR #2601, opened 2026-04-22
fix(guid): dedupe remote agents when merging detection and DB list

g199209:
## Summary

Deduplicate remote OpenClaw (and any remote) entries on the Guid **Agent pill bar** by merging the detected-agent list and the `remoteAgent.list` response without double-counting the same config.

## Problem (before this change)

On Guid, with **only one** remote agent configured in settings (e.g. OpenClaw), the pill bar showed **two** items:

1. A dimmed pill with key **`remote`** and label e.g. `OpenClaw` (no per-instance id in the selection key).
2. A selected pill with key **`remote:<uuid>`** and label e.g. `🤖 OpenClaw` (the real row from the database).

From the product perspective it looked like the user had added two agents; in reality it was one logical configuration rendered twice.

image

image

## Root cause

Two sources were concatenated in `useGuidAgentSelection` without normalizing `remote` rows:

- **`getAvailableAgents` / `DETECTED_AGENTS_SWR_KEY`**: comes from the main process `AgentRegistry`, which already loads remote rows from the DB, but the IPC response does **not** map `remoteAgentId` to the renderer’s `customAgentId`. The pill bar’s `getAgentKey()` only produces `remote:<id>` when `customAgentId` is set; otherwise it falls back to the bare string **`remote`**.
- **`remoteAgent.list`**: the authoritative list for Guid, mapped to `{ backend: 'remote', customAgentId: ra.id, avatar: ra.avatar, ... }`, which correctly yields keys **`remote:<uuid>`**.

The effect did `setAvailableAgents([...availableAgentsData, ...remoteAsAvailable])`, so the same real remote config appeared as both a **stub** (bare `remote`) and a **full** row (`remote:uuid`).

## What we changed

- Before appending the DB-backed remote rows, **filter out** `backend === 'remote'` entries from the detection/IPC list.
- Rely on **`remoteAgent.list`** for remote agents on Guid (id, name, avatar), consistent with the intent of the existing SWR key `remote-agents.list`.
- Add a **unit test** that simulates: detection returns `gemini` + a remote stub without `customAgentId`, and `remote.list` returns one row — assert a **single** `backend === 'remote'` entry and `getAgentKey` === `remote:<id>`.

After fix:

image

## Test plan

- [x] `bun run format`, `bun run lint`, `bunx tsc --noEmit`
- [x] `bun run test tests/unit/guidAgentSelection.dom.test.ts`
- [x] Manual: configure one remote OpenClaw, open Guid — only **one** pill for that agent; selection uses `remote:<id>` and avatar as before

### DIFF
--- a/src/renderer/pages/guid/hooks/useGuidAgentSelection.ts
+++ b/src/renderer/pages/guid/hooks/useGuidAgentSelection.ts
@@ -215,7 +215,11 @@ export const useGuidAgentSelection = ({
       customAgentId: ra.id,
       avatar: ra.avatar,
     }));
-    setAvailableAgents([...availableAgentsData, ...remoteAsAvailable]);
+    // Remote entries from getAvailableAgents lack customAgentId (IPC maps remoteAgentId
+    // only on the process side). remoteAgent.list is the source of truth for remotes
+    // (id, avatar) — drop remote stubs from detection to avoid duplicate pills.
+    const nonRemoteDetected = availableAgentsData.filter((a) => a.backend !== 'remote');
+    setAvailableAgents([...nonRemoteDetected, ...remoteAsAvailable]);
   }, [availableAgentsData, remoteAgentsData]);
 
   // Track whether the resetAssistant flag has been consumed so it only fires once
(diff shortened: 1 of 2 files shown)


piorpua:

PR #2615, opened 2026-04-22
fix(acp): surface real cause in crash banner + suppress false crash on graceful teardown

lixunqi12:
## Summary

- Enrich `DisconnectInfo` with `activePrompt` + `intentional` so the session layer can tell apart a real mid-turn crash, a graceful suspend, and an idle-time disconnect.
- `buildCrashMessage()` now returns `null` for intentional teardowns (suspend / user close), and for genuine crashes appends the disconnect `reason` (`pipe_close` / `connection_close` / `process_exit` / `process_close`), an "idle" vs "while a prompt was in flight" hint, and the last 400 chars of `stderr` â€” the user can now see a bun/node error or stack frame without leaving the chat.
- Delay the `stdout.close` observer in `ProcessAcpClient` by 200 ms so the `exit` event can land first on Windows and populate `child.exitCode`. Without this, every graceful shutdown showed `code: unknown` on Win32 because stdio close fires before exit there.

## Context

Follow-up to #2613. That PR logged diagnostic context from `ProcessAcpClient` but only via `console.warn`, which never reaches the daily log when the client runs inside a fork worker (Team leader / teammate sessions). This PR routes the same context through the existing `DisconnectInfo` â†’ `buildCrashMessage` path, which already terminates in the main process where `configureConsoleLog` can capture it â€” and, as a bonus, puts the context directly in the chat banner the user is staring at.

Reporting field comes from state `ProcessAcpClient` already tracks â€” no new process-level observations added.

## What does NOT change

- Error routing, teardown sequencing, and pending-request rejection behavior are unchanged.
- No new process events are added; the 4-signal detection (process_exit / process_close / pipe_close / connection_close) is preserved, only the payload it emits is enriched.
- `session/set_mode` / keepalive behavior is out of scope; idle-death mitigation is a separate PR.

## Test plan

- [x] `bunx vitest run tests/unit/process/acp/session/emitCrashSignal.test.ts` â€” 8/8 pass (3 new, 5 rewritten for new shape)
- [x] `bunx vitest run tests/integration/process/acp/session/AcpSession.lifecycle.test.ts tests/unit/process/acp/compat/AcpAgentV2.test.ts tests/unit/acpSessionOwnership.test.ts tests/unit/acpApplySessionMode.test.ts` â€” 77/77 pass
- [x] `bunx tsc --noEmit` â€” clean
- [x] `oxfmt --check` on touched files â€” clean
- [ ] Manual: suspend a session locally and confirm no crash banner appears (previously mislabeled as crash)
- [ ] Manual: kill a Claude ACP child mid-turn (Task manager) and confirm the banner now shows `reason: pipe_close` + stderr tail


piorpua:

PR #2626, opened 2026-04-23
fix(channels): skip invisible TMessage types to avoid overwriting final reply (#2625)

### LINKED ISSUE #2625: [Bug]: Telegram bot reply gets overwritten with "⏳ Processing..." placeholder after the real content arrives
### Platform

macOS (Apple Silicon)

### AionUi Version

1.9.19 (also reproducible on 1.9.20 — unchanged code path)

### Bug Description

When chatting with the AionUi Telegram bot, the reply sometimes renders the
correct assistant message on screen briefly and is then **overwritten back to
`⏳ Processing...`**. The desktop AionUi UI shows the full conversation
correctly — only the Telegram channel (and any other channel that relies on
`editMessage`) ends up stuck on the placeholder.

Observed backend: ACP / Claude (`backend: claude`), but the root cause applies
to any backend whose stream ends with a non-content event.

### Steps to Reproduce

1. Pair a Telegram bot via AionUi and configure an ACP/Claude (or Codex) backend.
2. Send a short message such as `你好` to the bot.
3. Watch the bot reply:
   - The placeholder `⏳ Thinking...` appears.
   - The real assistant text briefly shows up.
   - It then flips back to `⏳ Processing...` and stays there forever.

### Expected Behavior

The final `editMessage` should leave the actual assistant reply visible in
Telegram, matching what the desktop UI shows.

### Actual Behavior

Telegram stays stuck on `⏳ Processing...` even though the conversation on the
desktop app contains the full reply.

### Root Cause (analyzed)

`src/process/channels/gateway/ActionExecutor.ts` → `convertTMessageToOutgoing`
only handles `text / tips / tool_group / tool_call / acp_permission /
codex_permission`. Every other `TMessage` type (`agent_status`, `plan`,
`thinking`, `codex_tool_call`, `acp_tool_call`, `skill_suggest`,
`cron_trigger`, …) falls through to the `default` branch which returns a
hard-coded placeholder:

```ts
default:
  return {
    type: 'text',
    text: '⏳ Processing...',
    parseMode: 'HTML',
  };
```

In `handleChatMessage`, every streaming callback overwrites
`lastMessageContent`, and the **final** `editMessage(lastMsgId, finalMessage)`
uses that value. If the stream ends with any non-`content` event (very common
on ACP — e.g. a trailing `agent_status` after the assistant `text_delta`
content), the final edit replaces the real reply with the placeholder.

Because the error path inside `doEditMessage` and the final edit are both
wrapped in `try { ... } catch {}`, this is entirely silent — there is no log
entry on a successful-but-semantically-wrong overwrite.

### Additional Context

- The desktop UI is unaffected because it consumes `TMessage` directly.
- Network failures surface a different but related class of bug: transient
  `HttpError: Network request for 'editMessageText' failed!` is also swallowed
  by `try {} catch {}` around `doEditMessage`, leaving Telegram stuck on the
  last successfully edited intermediate state. Adding a retry/log for those
  would be a nice follow-up, but is **out of scope** for this bug report.

I've prepared a minimal fix (atomic, one-file) — see the accompanying PR.

81199000:
## Description

Fix Telegram/Lark bot replies getting overwritten with `⏳ Processing...`
after the real assistant content has already been streamed.

`convertTMessageToOutgoing` in `ActionExecutor.ts` routed every unrecognised
`TMessage` type (`agent_status`, `plan`, `thinking`, `codex_tool_call`,
`acp_tool_call`, `skill_suggest`, `cron_trigger`, …) into a hard-coded
`⏳ Processing...` placeholder via its `default` branch.

On channels that stream by editing a single placeholder message (Telegram,
Lark), `handleChatMessage` keeps `lastMessageContent` pointing at the last
converted outgoing message and uses it for the final `editMessage` call.
When the agent stream ends with one of those intermediate events — which is
the common case for ACP/Claude (a trailing `agent_status` follows the
assistant `content`) — the final edit silently replaces the real reply with
the placeholder string.

The desktop UI is unaffected because it consumes `TMessage` directly.

### Fix

- `convertTMessageToOutgoing` returns `null` from the `default` branch and
  widens its return type to `IUnifiedOutgoingMessage | null`.
- `handleChatMessage`'s streaming callback skips `null` conversions (they
  don't update `lastMessageContent`, don't drive any `editMessage`).

Tool progress (`tool_group`) and real assistant text (`text`) continue to
flow unchanged; the only events that stop being rendered to the channel
are purely internal/invisible ones that never carried meaningful content
for the end user anyway.

### Behaviour before / after

Before — message lands in Telegram and sticks on `⏳ Processing...`
(screenshots in issue #2625).

After — the final assistant reply stays visible; no more clobber.

## Related Issues

- Closes #2625

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tested on macOS (Apple Silicon, AionUi 1.9.19, Claude backend via ACP)
- [ ] Tested on Windows
- [ ] Tested on Linux
- [x] My code follows the project's code style guidelines (no-op for oxfmt on this file)
- [x] My changes generate no new warnings or errors

### Manual verification

1. Reproduced the bug: send `你好` to a paired Telegram bot with Claude/ACP
   backend. Observed the real reply flash on screen, then revert to
   `⏳ Processing...`.
2. Applied this patch, rebuilt/repackaged the asar, restarted AionUi.
3. Re-tested same prompt: reply renders once and stays rendered.
4. Verified tool-heavy prompts still show `⏳ tool_name` / `✅ tool_name`
   progress in Telegram (those flow through the `tool_group` case, not
   `default`).

## Additional Context

A separate concern is that both `doEditMessage` and the final
`context.editMessage(lastMsgId, finalMessage)` wrap calls in
`try { ... } catch {}`, which silently swallows `HttpError` from the
Telegram API. That explains a second symptom users occasionally hit
during transient network drops (reply never lands at all). Adding
logging + a single `sendMessage` fallback for the final edit would be a
reasonable follow-up, but is intentionally **out of scope** here to keep
this PR atomic per CONTRIBUTING rule #1.


piorpua:

PR #2630, opened 2026-04-23
fix(team): raise teammate inactivity watchdog from 60 s to 5 min

lixunqi12:
## Summary

In a 4-agent team running complex prompts (schematic review, long code audits), three of the four teammates would be flagged `"stopped responding after 60s without sending any update"` and cascade into `"process exited unexpectedly (code: unknown, signal: none)"` banners — despite the agents actually producing correct answers on the wire. Two independent things conspired:

1. **`WAKE_TIMEOUT_MS = 60 s` is too tight.** Claude Code and Codex routinely spend 1–3 minutes in an initial reasoning phase before their first streamed token on a hard prompt. The ACP-level prompt timeout already sits at 5 minutes; aligning the team watchdog to 5 minutes means a legitimately thinking agent does not lose to a stricter team-layer deadline.

2. **`handleResponseStream` gated the heartbeat reset on `agent.status === 'active'`.** If a stream event arrived while the slot was briefly `'pending'` (fresh wake) or `'idle'` (between back-to-back wakes), the reset was silently dropped and a stale timer from an earlier wake could fire before the current turn's first streamed token. `wakeTimeouts.has(slotId)` already means an armed timer for this slot exists, which is the only precondition we actually need; removing the status gate closes the false-kill race.

Symptom reproduction: a 4-agent team (leader + 3 teammates) is asked to audit a schematic hierarchy. Each teammate's first prompt takes ~90 s of internal reasoning before any streamed output. Under the old watchdog the 60 s timer fires mid-thought → `setStatus(slotId, 'failed')` → TeammateManager escalates → ACP session is torn down → chat shows the cryptic `process exited unexpectedly (code: unknown, signal: none)` red banner on three of four columns at once.

## Test plan

- [x] `bunx vitest run tests/unit/team-TeammateManager.test.ts` — **63 / 63 passing**
  - Updated the two existing watchdog tests to advance past the new 5-min deadline.
  - Added a regression test `resets the watchdog on streaming heartbeat even when slot status is not yet "active"` that flips the slot out of `'active'` after wake, emits a heartbeat at t=290 s, and asserts the t=300 s stale timer does not fire.
- [x] `bunx oxfmt --check src/process/team/TeammateManager.ts tests/unit/team-TeammateManager.test.ts` — clean
- [x] Full `bunx vitest run` — 4 491 / 4 495 passing; the 4 failing suites are pre-existing Windows symlink-permission tests on this host, unrelated to these changes (they fail identically on a clean `origin/main` checkout).
- [x] Manual smoke on the Windows desktop build: ran a 4-agent analog-automation team for ~20 min of back-to-back complex prompts; no false `process exited unexpectedly` banners. Before this PR the same workload produced one within the first two prompts.

### DIFF
--- a/src/process/team/TeammateManager.ts
+++ b/src/process/team/TeammateManager.ts
@@ -48,8 +48,18 @@ export class TeammateManager extends EventEmitter {
   /** Maps slotId → original name before rename, for "formerly: X" hints in prompts */
   private readonly renamedAgents = new Map<string, string>();
 
-  /** Maximum time (ms) to wait for a turnCompleted event before force-releasing a wake */
-  private static readonly WAKE_TIMEOUT_MS = 60 * 1000;
+  /**
+   * Maximum time (ms) a teammate may go silent between streaming updates before
+   * the wake is escalated to {@link handleInactivityTimeout}. Needs to be long
+   * enough to cover an agent's initial thinking phase on a complex prompt —
+   * Claude Code and Codex regularly spend 1–3 minutes reasoning before their
+   * first streamed token on hard questions (schematic hierarchies, long
+   * reviews, architectural decisions). The old 60 s value killed legitimate
+   * turns mid-thought; 5 minutes aligns with the ACP-level prompt timeout so
+   * a truly stalled agent still gets flagged, but normal long thinking does
+   * not.
+   */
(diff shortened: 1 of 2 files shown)


piorpua:

PR #2639, opened 2026-04-24
fix(team): bind updateTask SQL to resolved id for short prefixes

g199209:
## Summary

- `SqliteTeamRepository.updateTask()` was binding the raw input `id` (which may be a short prefix like `"12345678"`) to the `WHERE id = ?` SQL clause, causing the UPDATE to silently match nothing when a short prefix was passed
- Fixed by using `current.id` (the full UUID resolved by `findTaskById`) as the SQL bind parameter
- Added integration test covering the short-prefix → full-UUID update path

## Test plan

- [x] Run `bun run test` — new test `updateTask > updates the full task row when called with a short id prefix` should pass
- [x] Verify that task updates via short prefix IDs are now correctly persisted in the database


piorpua:

PR #2641, opened 2026-04-24
fix(team): team_task_update omit owner when unset; unblock resolved id

g199209:
## Summary

- `team_task_update` tool was spreading `owner: undefined` into the task patch when the caller only updated status, silently unassigning the task in the repository layer
- Fixed by only including fields the caller actually specified (omit `owner` if not provided or blank)
- Use the resolved task id returned from `taskManager.update()` for `checkUnblocks()` calls, so short-id lookups work correctly

## Test plan

- [x] Verify status-only updates no longer overwrite the existing task owner
- [x] Verify blank owner strings are treated as omitted
- [x] Verify trimmed owner is forwarded when provided
- [x] Verify `checkUnblocks` is called with the full resolved UUID, not the short id passed by caller


piorpua:

PR #2643, opened 2026-04-24
fix(team): unassign orphan tasks after teammate removal

g199209:
## Summary

- When a teammate is removed, any `pending` or `in_progress` tasks they owned were left with a ghost owner, causing the task board to show stale assignments and silently stalling work.
- This fix unassigns those orphaned tasks and sends a notification to the leader so they can explicitly reassign or drop the work.
- Extends the `onAgentRemoved` callback signature to pass the removed `TeamAgent` object, enabling the cleanup without an extra lookup.

## Test plan

- [x] Remove a teammate who owns pending/in-progress tasks → verify tasks become unassigned and the leader receives a mailbox message listing them
- [x] Remove a teammate with no active tasks → verify no notification is sent and `updateTask` is not called
- [x] Remove a teammate whose tasks are all `completed` → verify those tasks are untouched
- [x] Unit tests added in `team-TeamSession` and `team-TeammateManager` suites covering all three scenarios


piorpua:

PR #2644, opened 2026-04-24
fix(team): keep task owners in sync when renaming teammates

g199209:
## Summary

- When a teammate is renamed, any tasks they own retain the old name as owner, causing the task board to show stale owners that no longer match `team_members`
- `TeamSession.renameAgent` now captures the previous name before delegating to `TeammateManager`, then calls `persistRename` which updates the DB and rewrites all owned task `owner` fields to the new name
- `TeamSessionService.renameAgent` handles the offline (no active session) case identically via `syncTaskOwnersForRename`
- Both code paths are no-ops when the trimmed new name equals the old name, avoiding unnecessary DB writes

## Test plan

- [ ] Rename an active teammate — verify tasks previously owned by them now show the new name in `team_task_list`
- [ ] Rename a teammate while the team session is offline — verify same task owner sync occurs via `TeamSessionService`
- [ ] Rename to the same name (with/without surrounding whitespace) — verify no task update queries are issued
- [ ] Unit tests added for `TeamSession.renameAgent()` (2 cases) and `TeamSessionService.renameAgent` offline path (1 case)


piorpua:

PR #2689, opened 2026-04-28
fix(acp): resolve Windows shell command path for opencode startup

zhangxingyu941:
## Summary
- resolve bare ACP backend commands to concrete executable paths before Windows shell spawn
- avoid launching quoted bare commands such as `"opencode"` through `cmd.exe`
- ensure backends such as `opencode` start from the intended executable path
- preserve existing behavior by falling back to the original command when `where` does not return a valid executable path

## Background
On Windows, ACP backends are started through `cmd.exe` so AionUi can prepend:

```bash
chcp 65001 >nul &&
```

This is needed to force UTF-8 console output and avoid garbled stdout/stderr on systems that default to a non-UTF-8 code page.

Before this change, a bare backend command such as `opencode` was converted into:

```bash
chcp 65001 >nul && "opencode"
```

## Problem
Launching `opencode` this way was unstable on Windows.

The main issue is that `cmd.exe` handles a quoted bare command name such as `"opencode"` differently from the unquoted command `opencode`.

In this case, quoting the bare command caused OpenCode's internal path inference to resolve the wrong module entry path. The observed failure looked like this:

```text
C:\Users\lenovo>chcp 65001 >nul && "opencode"
node:internal/modules/cjs/loader:1404
  throw err;
  ^

Error: Cannot find module 'C:\Users\lenovo\global\5\.pnpm\opencode-ai@1.2.15\node_modules\opencode-ai\bin\opencode'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1401:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1057:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1062:22)
    at Function._load (node:internal/modules/cjs/loader:1211:37)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v22.17.0
```

The command was available on `PATH`:

```text
C:\Users\lenovo>where opencode
C:\Users\lenovo\AppData\Local\pnpm\opencode
C:\Users\lenovo\AppData\Local\pnpm\opencode.CMD
```

So detection succeeded, but startup failed because the final shell command launched the quoted bare command form.

## Root Cause
The Windows startup path always wrapped the configured `cliPath` in quotes:

```bash
chcp 65001 >nul && "opencode"
```

This is correct for real filesystem paths with spaces, such as:

```text
C:\Program Files\agent\agent.exe
```

However, it is problematic for bare command names like `opencode`.

For bare commands, quoting changes how `cmd.exe` performs command resolution. With OpenCode, that altered resolution path caused its Node entrypoint logic to infer a broken module path, resulting in `MODULE_NOT_FOUND`.

In short:
- AionUi generated a shell command using the quoted bare name `"opencode"`
- `cmd.exe` resolved that form differently from `opencode`
- OpenCode's internal module path inference broke
- backend launch failed with `MODULE_NOT_FOUND`

## Fix
This change resolves bare Windows commands before building the final shell command.

The new behavior is:
- detect whether the configured backend command is a bare command such as `opencode`
- run `where <command>` to resolve the actual executable path from the current `PATH`
- prefer a valid executable candidate such as `opencode.CMD`
- accept only valid Windows executable paths from `where`
- use that resolved path when constructing the final `spawnCommand`
- if resolution fails or returns unexpected output, safely fall back to the original command

So instead of launching a quoted bare command:

```bash
chcp 65001 >nul && "opencode"
```

AionUi now launches the resolved executable path, for example:

```bash
chcp 65001 >nul && "C:\Users\lenovo\AppData\Local\pnpm\opencode.CMD"
```

This keeps the existing UTF-8 shell startup behavior while avoiding the problematic quoted bare-command form.

## Result
After this change, Windows ACP startup no longer depends on `cmd.exe` resolving a quoted bare command like `"opencode"`.

AionUi resolves the executable path up front and uses that path directly, which avoids the OpenCode startup failure shown above and allows OpenCode connections to start correctly.

## Verification
- `bunx vitest run tests/unit/acpConnectors.test.ts`


piorpua:

PR #2710, opened 2026-04-30
 fix(aionrs): resolve binary path and expose startup errors in development mode

safesummer:
## Summary

  When starting the dev server with `bun run webui:prod:remote` (Renderer + Node process without Electron shell), the ai
  onrs agent fails to initialize because `process.resourcesPath` points to `electron/dist/resources`, which does not con
  tain the bundled binary.

  This PR fixes three related issues discovered during local development:

  1. **Add dev-mode binary path fallback** in `binaryResolver.ts` — fall back to `process.cwd()` and parent directory wh
  en the bundled path under `resourcesPath` is missing
  2. **Expose startup errors** in `AionrsManager.ts` — replace silent `.catch(() => {})` with `mainError()` so failures
  are visible in logs
  3. **Fix `runtimeKey` scope** — move declaration to function top so fallback paths can reference it

  ## Root cause

  In dev mode (`webui:prod:remote`), `process.resourcesPath` resolves to Electron's internal resources directory, not th
  e project root. The original code only checked this path and then fell back to PATH, never looking in the actual proje
  ct tree where `resources/bundled-aionrs/` lives.

  ## Test plan

  - [x] Verified aionrs starts successfully in local dev mode (`bun run webui:prod:remote`)
  - [x] Verified two-round conversation works ("你好" and "1+1等于几" both receive responses)
  - [x] All existing tests pass (`bunx vitest run`)
  - [x] Lint and type-check pass (`bun run lint`, `bunx tsc --noEmit`)
  - [x] Added unit tests for `binaryResolver.ts` covering all resolution paths (bundled, cwd, parent cwd, PATH, not-foun
  d)
  
BTW I also found something wrong when I tried to upload files at linux serve releases , I'll figure it out later.(this pr above was by kimi -vibe coding ,but I tied many times ,please check it out  carefully ,thks!)


safesummer:
通过vibe coding（虽然开源作者们都不太喜欢它的质量），但它确实简单实现了多租户账号分配、分隔对话功能（尽管有潜在的bug和未考虑到的问题），希望以后的版本可以支持。这部分我就不敢提交了hh

PR #2717, opened 2026-04-30
Cumulative Aion CLI/UI fixes + cron multi-binding + slash commands on v1.9.23

cheats1314:
Cumulative customizations on top of `v1.9.23`, organized by area.

### White screen on every Aion CLI turn — `Cannot read properties of undefined (reading 'conversation_id')`
`transformMessage` declared a `TMessage` return type but implicitly returned `undefined` for `request_trace`, `info`, `system`, `acp_model_info`, `acp_context_usage`, `available_commands` and several other types. The aionrs hook's `default` switch case forwarded that `undefined` to `useAddOrUpdateMessage`, which queued `{ message: undefined }` and then crashed inside `flush()` at `item.message.conversation_id` — bubbling to the renderer's `ErrorBoundary` on every AI turn (because `request_trace` fires on every turn).

- Mark `transformMessage` return type as `TMessage | undefined` to match real behaviour.
- Guard `useAddOrUpdateMessage` to drop `undefined` at the entry point so the message list never holds undefined items.
- Add `!data` / `!msg` guards to the confirmation, skill-generator, model-health-check, team-permission and team-badge listeners that read `conversation_id` on potentially-undefined IPC payloads.
- Add the same guards across the per-platform response stream hooks (acp/aionrs/gemini/openclaw/nanobot/remote).

### Aion CLI permission mode UI not refreshing after switch
`AgentModeSelector`'s single `useEffect([initialMode, modes, defaultMode])` reverted `currentMode` on every backend `config_changed` event because `mergeWithCapabilities` returns a new array reference; the user's just-made selection got overwritten by the still-stale `initialMode` prop, so the UI only caught up after the conversation was navigated away and back. Split into two effects with separate responsibilities — parent-driven `initialMode` sync vs. modes-list re-validation — so user selections survive capability refreshes.

### Aion CLI / aionrs startup hardening
- SQLite create-conversation failures no longer silently ignored.
- aionrs sessions missing `model` give an explicit error instead of failing silently.
- Conversation route shows a fallback when the route is missing/corrupt instead of white-screening.
- aionrs panel correctly forwards `workspacePath`.
- Default model selection skips unavailable providers; `gemini-with-google-auth` is no longer eligible as the aionrs default.

### Renderer `ErrorBoundary`
Added in `src/renderer/main.tsx` so transient renderer failures show a recovery card instead of blanking the whole app.

## Cron / Scheduled tasks
- Multi-conversation bindings per cron job (one job ↔ many conversations).
- Cron job ↔ agent config decoupled; UI binding + remote session entry redesigned.
- Custom interval scheduled tasks.
- Refactored `CronJobManager` UI; cron-trigger card accessibility.

## Layout / Mobile / Resizable sider
- New `useResizableSider` hook for the workspace ↔ chat splitter; tests added.
- Team-mode member-to-member separator drag (each slot keeps a min width of 280px).
- Mobile workspace drawer narrowed.
- Heavier scrollbar styling.

## Slash commands & sendbox UX
- Claude Code-style `/rewind` turn picker with keyboard navigation, numeric shortcuts, `0` for latest target, and composer refocus after rewind.
- `/undo` unified with the `/rewind` picker.
- Built-in `/clear` and `/help` slashes for ACP conversations; ACP child-process shutdown is awaited before resolving so there are no clear/rewind races.
- `/model` slash entry.
- ArrowUp now uses visual-line caret detection in the browser (hidden mirror div), so it no longer jumps to the start of multi-line input.

## Message history loading
- Apply cached history before first paint to avoid a blank message pane on conversation switch.
- Load latest messages first on switch; staged loading; harden order parameter.
- Retry on conversation re-focus; keep cache fresh during local updates.
- Lightweight performance instrumentation (DB load latency, payload size, render readiness).

## Accessibility polish
Improved a11y on the message navigation rail, scroll-to-bottom button, selected-text reply button, scheduled-task card, and thinking details.

## Tests
New / expanded unit + DOM tests for: layout resize, mobile layout, `useResizableSider`, sendbox history (ArrowUp visual-line caret), cron store/service, message hooks, message list loading, platform sendboxes, selection reply button, ACP history e2e.

## Notes for reviewers

- The branch is based off `v1.9.22`; the final commit aligns it to `v1.9.23` (package version + WeChat QR refresh) so this PR's diff stays focused on the actual customizations.
- **`.github/workflows/release-distribute.yml` shows up as deleted only because the GitHub OAuth token used to push this branch lacks the `workflow` scope and therefore could not carry the file forward when this branch was created from an older fork base. It is not an intentional removal — please ignore that file in review.**
- The earliest user-facing changes were squashed into one large baseline commit (`Update to v1.9.22 with complete interaction fixes`) when migrating from 1.9.21; subsequent commits in this branch are the focused fixes described above.

## Test plan

- [x] `npx tsc --noEmit` passes.
- [x] Targeted DOM tests pass (`sendboxHistory`, `useResizableSider`, `Layout.resize`, `useGeminiMessage`, `platformSendBoxes`, `GuidActionRow`, `CreateTaskDialog`).
- [x] Manual verification on a fresh Linux `.deb` build:
  - Aion CLI no longer white-screens after AI replies (root-cause fix verified).
  - Aion CLI permission mode UI updates immediately on switch (no more navigate-away-and-back).
  - ArrowUp behaves correctly in multi-line input.
  - Team-mode member separator drag works.


piorpua:

Issue #1896, opened 2026-03-30
[Feature]: Support Proxy Settings


ScathonLin:
### Feature Description

The network is not reachable in my company PC because of the ACL rules, if I want to visit the internet, I can set the proxy with my domain username and password, but AionUI doesn't have the entry to do this;

### Problem Statement

The network is not reachable in my company PC because of the ACL rules, if I want to visit the internet, I can set the proxy with my domain username and password, but AionUI doesn't have the entry to do this;

### Proposed Solution

Add a entry in the settings for user to setup proxy settings;

### Feature Category

UI/UX Improvement

z91300:
+1

m0leynik:
+1
Same situation

PR #2773, opened 2026-05-07
feat(proxy): add global proxy settings

cdxiaodong:
## Summary
- add a global proxy setting in System settings and persist it in shared config storage
- propagate the proxy to Gemini and Aionrs agent startup paths
- support proxy environment injection for aionrs and add unit coverage for the new bridge and env builder behavior

## Test plan
- [x] `bun run format`
- [x] `bun run lint`
- [x] `bunx tsc --noEmit`
- [x] `bun run i18n:types`
- [x] `node scripts/check-i18n.js`
- [x] `bunx vitest run tests/unit/systemSettingsBridge.test.ts tests/unit/process/agent/aionrsEnvBuilder.test.ts`


cdxiaodong:
Closing this duplicate in favor of #2318, which already has green checks and prior review context for issue #1896.

Issue #2757, opened 2026-05-06
[Bug]: Key rotation not working


liberom:
### Platform

Windows

### AionUi Version

1.9.25

### Bug Description

So far I've only seen this for Openrouter, but I inserted multiple API keys (tried both: comma separated and one per line), but I keep getting "Provider error: Connection error: Invalid authorization header: failed to parse header value". I tried asking Gemini for solutions, but none worked (insert 'Bearer ' or just ' ' before the key, etc).
Aion detects that there are multiple keys, but it's not handling the rotation properly.

### Steps to Reproduce

- Add multiple keys
- Try chatting with a model
- Get the error

### Expected Behavior

No error should appear. The system should handle key rotation one at a time.

### Actual Behavior

What I believe is happening is that all keys are being sent at the same time, so I am unable to chat with any Openrouter model if I use multiple keys.

zhangyapu1:
me too

PR #2785, opened 2026-05-08
fix(api): honor retry-after for rotating clients

### LINKED ISSUE #2782: RotatingApiClient.executeWithRetry — 429 retries ignore Retry-After, key rotation spreads pressure during provider cooldown
## Finding

`RotatingApiClient.executeWithRetry()` treats
401, 429, 503, and 5xx responses as the same
retry class — a single generic retry path with
fixed linear delay and optional key rotation.

On 429, the client does not inspect `Retry-After`
or any provider cooldown signal before retrying.

## What happens under rate limiting

When a provider returns 429 with `Retry-After`:

1. The client retries with fixed linear delay —
   before the provider cooldown expires
2. If multiple API keys are configured, key
   rotation occurs during the cooldown window
3. Rotated keys hit the same provider cooldown
   and fail — spreading transient pressure across
   keys instead of waiting for reset

Key rotation is designed to help with rate limits.
Under this pattern it makes them worse.

## The consequence

At scale, treating 429 cooldowns as generic
retryable errors and rotating keys can create
premature retry traffic and spread throttling
pressure across keys during provider rate-limit
windows.

Retries fired inside a provider cooldown window
are guaranteed to fail. They burn retry budget,
waste API key quota, and extend actual recovery
time.

## The correct behavior

On 429 with `Retry-After`:
- Read the header value
- Wait at least that duration before retrying
- Do not rotate keys during active provider cooldown
- Fall back to exponential backoff only when
  `Retry-After` is absent

## Related pattern

`retry-after-ignored-under-concurrency`

Corpus reference: https://github.com/SirBrenton/pitstop-truth

cdxiaodong:
## Summary
- honor `Retry-After` on 429 responses before retrying rotating API client calls
- avoid rotating API keys while a provider cooldown is active
- fall back to exponential backoff when 429 responses do not include `Retry-After`
- add regression coverage for cooldown retries, fallback backoff, and non-429 key rotation

## Test plan
- [x] `bunx vitest run tests/unit/common/rotatingApiClient.test.ts`
- [x] `bunx tsc --noEmit`

Closes #2782
Related to #2757


piorpua:

PR #2786, opened 2026-05-08
fix(model): update minimax default endpoint

### LINKED ISSUE #2776: [Bug]: MiniMax provider uses outdated minimaxi.com endpoint and existing configs are not migrated
### Platform

macOS (Apple Silicon)

### AionUi Version

v1.9.24

### Bug Description

 When configuring the built-in MiniMax provider in AionUi, the app uses https://api.minimaxi.com/v1 as the base URL. That endpoint appears to be outdated/incorrect and should be https://api.minimax.io/v1. Therefore minimax provider not working, not responding.

There also seems to be a migration gap: even if the app default is corrected in a newer build, existing users may still keep the old minimaxi.com URL because the saved provider config is persisted and not automatically updated.

### Steps to Reproduce

1. Open AionUi on macOS.
2. Go to Settings -> Model.
3. Create or inspect the built-in MiniMax provider configuration.
4. Check the default/base URL shown for MiniMax.
5. If you already had MiniMax configured before, restart AionUi and inspect the saved provider entry again.

### Expected Behavior

The MiniMax provider should default to:  https://api.minimax.io/v1
If an older persisted config still uses:  https://api.minimaxi.com/v1

AionUi should migrate that saved value automatically to https://api.minimax.io/v1.

### Actual Behavior

AionUi shows/saves the MiniMax provider base URL as: https://api.minimaxi.com/v1

Existing saved MiniMax configs continue to use that old value, so users may still see the wrong endpoint even after updating the app unless they manually edit their config.

### Additional Context

**Environment:**
  - Platform: macOS (Apple Silicon)
  - AionUi version: v1.9.24

  **What I found while debugging locally:**
  - The installed app bundle contained a MiniMax default provider entry using https://api.minimaxi.com/v1.
  - My saved user config also had a persisted MiniMax entry still pointing to https://api.minimaxi.com/v1.
  - Updating only the app bundle default was not enough for existing users, because the persisted config continued to override the default.
  - After manually changing the saved MiniMax config to https://api.minimax.io/v1, the issue was resolved locally.

  **Suggested fix:**
  1. Update the MiniMax default provider URL to https://api.minimax.io/v1.
  2. Add a startup migration for persisted model.config entries where name === "MiniMax" and baseUrl === "https://api.minimaxi.com/v1".

cdxiaodong:
## Summary
- update the built-in MiniMax provider URL from `api.minimaxi.com` to `api.minimax.io`
- keep MiniMax model-list detection covered by the bridge unit test using the new endpoint

## Test plan
- [x] `bunx vitest run tests/unit/bridge/modelBridge.test.ts`
- [x] `bunx tsc --noEmit`

Closes #2776


piorpua:

PR #2788, opened 2026-05-08
fix(acp): preserve custom agent launch config

### LINKED ISSUE #2775: [Windows] Custom ACP agents fail to start: `--experimental-acp` injected + custom agent config never loaded
## Summary

On Windows (v1.9.25), creating a custom ACP agent through the UI (Settings → Agents → Local Agents → Detect Custom Agent) results in a working **Test Connection** but every real conversation fails with `Session failed to start` / `process exited unexpectedly`.

Two distinct bugs combine to cause this. Both have been confirmed by patching `app.asar` locally and observing AionUi behavior with debug logging injected.

---

## Repro

1. Have any CLI that speaks ACP but does NOT support `--experimental-acp` as a flag of the binary itself. Easiest reproducer: anything wrapped in `docker exec` (because `docker` itself rejects unknown flags before reaching the container command). Example used here: `docker exec -i <container> <my-acp-binary> acp`.
2. Settings → Agents → Local Agents → Detect Custom Agent
   - Display Name: `My Custom Agent`
   - Command: `docker`
   - Arguments: `exec -i <container> <my-acp-binary> acp`
   - Env: (any, e.g. `PYTHONUNBUFFERED=1` if the container runs Python)
3. Click **Test Connection** → ✅ "Connection successful! CLI exists and ACP protocol is working."
4. Save, start a new conversation with this agent, send any prompt.
5. **Expected**: the agent responds.
   **Actual**: 4 retries of `process exited unexpectedly (code: 125, signal: none)` then `Agent disconnected (connection_close)`. AionUi log shows `[ACP custom STDERR]: unknown flag: --experimental-acp / Usage: docker [OPTIONS] COMMAND [ARG...]`.

This same class of failure will happen for any custom binary that does not happen to accept `--experimental-acp` as a flag.

---

## Bug 1 — `--experimental-acp` is auto-injected for custom agents

**File**: `src/process/agent/acp/acpConnectors.ts` (compiled to `out/main/index.js:13247`)

```ts
const effectiveAcpArgs = acpArgs === void 0 ? ['--experimental-acp'] : acpArgs;
```

The default `['--experimental-acp']` makes sense for `claude`, but is fatal for any other binary (here `docker` itself rejects the unknown flag → exit 125).

**Important context**: every built-in backend in `ACP_BACKENDS_ALL` already declares its own `acpArgs` explicitly (`['--acp']`, `['acp']`, `['exec', '--output-format', 'acp']`, `[]`, etc.). **None of them rely on the default**. So the default `['--experimental-acp']` only ever fires for custom agents whose `acpArgs` happens to be `undefined` — which leads us to bug 2.

---

## Bug 2 — Custom-agent config is fetched from the wrong storage key

**File**: `src/process/task/AcpAgentManager.ts` (compiled to `out/main/index.js:40397`)

```ts
async resolveCustomAgentCliConfig(data) {
  const customAgents = await ProcessConfig.get("assistants");
  let customAgentConfig = customAgents?.find(
    (agent) => agent.id === data.customAgentId
  );
  // ...
  if (!customAgentConfig?.defaultCliPath) {
    return { cliPath: data.cliPath };  // ← falls through, no customArgs, no customEnv
  }
  return {
    cliPath: customAgentConfig.defaultCliPath.trim(),
    customArgs: customAgentConfig.acpArgs,
    customEnv: customAgentConfig.env,
  };
}
```

The lookup reads `ProcessConfig.get("assistants")` — but on disk that key only ever contains the **built-in assistant templates** (all `builtin-*` ids: `builtin-cowork`, `builtin-ppt-creator`, etc.). User-saved custom agents are persisted under a different key, **`acp.customAgents`** — which is exactly what `getDetectedCustomAgents()` reads at `out/main/index.js:12112`:

```ts
const customAgents = await ProcessConfig.get("acp.customAgents");  // ← detection uses this
```

So detection sees the agent (it shows up in the UI, Test Connection works because that path uses `customAgents` from a different source), but resolution doesn't, falls through to `{ cliPath: data.cliPath }`, returns no args / no env, and bug 1 then defaults `acpArgs` to `['--experimental-acp']` → docker spawned with one unknown flag → 125.

Confirmed via injected debug logging in the patched binary:

```
resolveCustom rawId="<local-uuid>" found=false 
availableIds=["builtin-story-roleplay","builtin-beautiful-mermaid",...,"builtin-word-creator"]
```

(The user's saved agent has a generated UUID, but only `builtin-*` ids are present in the `assistants` key.)

The on-disk config storage clearly shows the agent under the right key:
```json
"acp.customAgents": [{
  "id": "<local-uuid>",
  "name": "My Custom Agent",
  "defaultCliPath": "docker",
  "acpArgs": ["exec","-i","<container>","<my-acp-binary>","acp"],
  "env": {"PYTHONUNBUFFERED": "1"}
}]
```

---

## Suggested fix

**Bug 2 (root cause)** — patch `resolveCustomAgentCliConfig` to read from the same source as detection, e.g. merge both keys for safety:

```ts
const builtinAssistants = (await ProcessConfig.get("assistants")) || [];
const userCustomAgents = (await ProcessConfig.get("acp.customAgents")) || [];
const customAgents = [...builtinAssistants, ...userCustomAgents];
```

(Bonus: also accept ids prefixed with `custom:` since the UI persists `guid.lastSelectedAgent` as `"custom:<id>"`.)

**Bug 1 (defense in depth)** — change the default to `[]` instead of `['--experimental-acp']`:

```ts
const effectiveAcpArgs = acpArgs === void 0 ? [] : acpArgs;
```

This is safe because every built-in backend already declares its own `acpArgs` explicitly — only mis-resolved custom agents would ever hit the default, and `[]` fails loudly (clear "no command" error) instead of silently injecting a flag that breaks arbitrary binaries.

---

## Environment

- AionUi v1.9.25 (Windows installer `AionUi-1.9.25-win-x64.exe`)
- Windows 10 Enterprise 19045
- Custom agent: any CLI that speaks ACP over stdio but does not accept `--experimental-acp` as a flag of the main binary. Reproduced here with a custom CLI agent wrapped via `docker exec -i` (stdio JSON-RPC v1).
- Both fixes verified locally by patching `app.asar` — the custom agent now connects and answers prompts inside AionUi alongside Claude / Gemini / Codex.

Happy to send a PR if a maintainer confirms the suggested approach.

cdxiaodong:
## Summary
- resolve user custom ACP agents from `acp.customAgents` in addition to built-in assistants
- accept persisted `custom:<id>` custom-agent identifiers when resolving launch config
- stop injecting `--experimental-acp` when ACP args are omitted for generic custom commands
- add regression coverage for omitted ACP args

## Test plan
- [x] `bun run format`
- [x] `bunx vitest run tests/unit/acpConnectors.test.ts`
- [x] `bunx tsc --noEmit`

Closes #2775


piorpua:

PR #2793, opened 2026-05-08
fix(preview): keep sidebar state on team switch with open preview

eryk:
## Summary

- Lift `usePreviewAutoCollapse`'s transition memory (`previousPreviewOpen`, `previousWorkspaceCollapsed`, `previousSiderCollapsed`) from local refs in the hook to a `MutableRefObject` owned by `PreviewProvider`, exposed through `PreviewContextValue.autoCollapseMemoryRef`.
- Add unit tests for `usePreviewAutoCollapse` covering the team-switch remount path that was previously uncovered.

## Bug

Switching teams while the preview panel is open and the sidebar is open causes the sidebar to be force-collapsed again — even though the user did not just open the preview.

## Root cause

`src/renderer/pages/team/index.tsx:15` uses `key={team.id}` so `TeamPage`/`ChatLayout` fully remounts when the team id changes. `usePreviewAutoCollapse` previously held its three transition refs (`previousPreviewOpenRef`, `previousWorkspaceCollapsedRef`, `previousSiderCollapsedRef`) inside the hook, so they were reset to their initial values on every remount.

The root-level `PreviewProvider` (mounted in `src/renderer/main.tsx`) is **not** under that `key` boundary, so its `isPreviewOpen` state stayed `true` across the remount. The newly mounted hook saw `isPreviewOpen=true && previousPreviewOpen=false` and treated it as a fresh "preview just opened" transition, running the force-collapse branch a second time and stomping the user's sidebar state.

## Fix

Move the transition memory to a `MutableRefObject<PreviewAutoCollapseMemory>` owned by `PreviewProvider`. Because `PreviewProvider` is mounted at the app root (above the team-page `key` boundary), the ref survives `ChatLayout` remounts. `usePreviewAutoCollapse` now reads/writes this shared memory via `usePreviewContext()` instead of using its own `useRef`s. Behavior on the open/close paths is otherwise unchanged.

The three refs are aggregated into a single object so the public context surface only grows by one field.

## Files changed

- `src/renderer/pages/conversation/Preview/context/PreviewContext.tsx` — new `PreviewAutoCollapseMemory` interface; `PreviewContextValue` gains `autoCollapseMemoryRef`; `PreviewProvider` initializes the ref and exposes it.
- `src/renderer/pages/conversation/Preview/context/index.ts` — re-export the new type.
- `src/renderer/pages/conversation/hooks/usePreviewAutoCollapse.ts` — drop the three internal `useRef`s; read `autoCollapseMemoryRef` from `usePreviewContext()`; equivalent logic against the shared memory.
- `tests/unit/usePreviewAutoCollapse.dom.test.ts` — new file, 5 cases (see Test plan).

## Untouched (intentionally)

- `src/renderer/pages/team/index.tsx` — `key={team.id}` is required for other reasons; not removed.
- `src/renderer/components/layout/Layout.tsx` — `siderCollapsed` state unchanged.
- `src/renderer/pages/conversation/components/ChatLayout/index.tsx` — call site unchanged.
- `PreviewContext` `isOpen` non-restoration from localStorage — unchanged.
- Mobile early-return path in `usePreviewAutoCollapse` — unchanged.

## Test plan

- [x] Acceptance — team switch with preview open + sidebar open → sidebar stays open (covered by `does not re-trigger force-collapse on remount when preview is still open`).
- [x] Acceptance — team switch with preview open + sidebar collapsed → sidebar stays collapsed (same test path, since the hook never calls setters under that condition).
- [x] Acceptance — team switch with preview closed + sidebar open/collapsed → sidebar unchanged (closed branch never runs because `previousPreviewOpen` is false; not regressed by this PR).
- [x] Same-team — open preview → sidebar auto-collapses (`force-collapses sidebar and workspace when preview opens for the first time`).
- [x] Same-team — close preview → sidebar restored (`restores sidebar to its prior state when preview closes`).
- [x] Regression — team switch then close preview → sidebar restored to the pre-open state captured under the previous team (`restores sidebar to its pre-open state after team switch when preview is closed`).
- [x] Mobile path unchanged (`skips collapse logic on non-desktop layouts`).
- [x] `bun run test` — 4533 passed (+1 new).
- [x] `bun run format` — clean.
- [x] `bun run lint` — 0 errors (warnings unchanged from main).
- [x] `bunx tsc --noEmit` — no new errors in changed files.


piorpua:

PR #2890, opened 2026-05-12
fix(acp): drop session/load replay events on resume

### LINKED ISSUE #2887: [Bug]: Session resume duplicates conversation history in SQLite — replay events inserted as new rows on every restart (Claude Code backend)
### Platform

Windows

### AionUi Version

1.9.25

### Bug Description

When an AionUi conversation is reopened after the app is fully closed and restarted, the next user message triggers AionUi to insert the prior conversation turns into SQLite as new rows with new `msg_id` values, rather than reusing the existing rows. The `messages` table grows on each restart by approximately the number of prior Lead-side turns, and consecutive replay events are concatenated by the stream-chunk buffering, producing rows whose content spans multiple original turns.

The underlying ACP `session/load` mechanism appears to work correctly at the model layer (the resumed agent has full prior context and responses are correct). The bug is in how AionUi's stream-event handler processes the replay events Claude Code emits during `session/load`: the events flow through `onStreamEvent` → `transformMessage` → `addOrUpdateMessage`, and because `transformMessage` does not preserve the original session message UUIDs, `addOrUpdateMessage`'s upsert semantics inserts duplicates instead of deduplicating.

The user-visible symptom is that on the first post-restart message, the chat pane re-renders the prior conversation before showing the new response. The underlying problem is more substantive than the cosmetic appearance suggests: the SQLite database bloats proportionally to (prior conversation length) × (number of restarts), and the merged-row content corruption means stored turn boundaries no longer match the original conversation flow.

### Steps to Reproduce

1. Create a Team in AionUi with Claude Code as the Lead. Set a workspace.
2. Have an extended back-and-forth conversation with the Lead until you have ~10-20 substantive Lead-side text turns plus tool calls (any topic; the content doesn't matter for the bug).
3. Note the current row count for the conversation:
     SELECT COUNT(*) FROM messages WHERE conversation_id = '<id>';
4. Fully quit AionUi (system tray → Quit; verify no AionUi process is running via Task Manager).
5. Wait ~30 seconds. Reopen AionUi.
6. Navigate to the same Team. The conversation history loads correctly from SQLite (this part works).
7. Send any prompt to the Lead — a trivial recall-only request is sufficient, e.g., a request to summarize what was discussed previously, with explicit instruction not to read files or invoke tools.
8. Observe in the UI: the Lead's prior turns re-appear before the new response is generated.
9. Re-run the row count query and compare with step 3.

### Expected Behavior

A recall-only prompt that requires no tools and no dispatches should add ~2 rows to the `messages` table:
  - 1 row for the user's prompt
  - 1 row for the Lead's new text response

A restart by itself should add 0 rows. The previously stored conversation rows should remain the same rows after restart — the resumed session should re-use them via msg_id deduplication, not duplicate them.

In a captured run with 14 Lead-side text turns before restart, the post-restart row count should have been approximately:
  (prior count) + 2 rows for the new exchange.

### Actual Behavior

After step 7, the observed growth was ~9 new rows in addition to the legitimate new response. The new rows have brand-new `msg_id` UUIDs (not present in the pre-restart rows) and their content is replays of prior Lead-side turns with consecutive original turns merged into single rows.

Captured evidence (timestamps via `created_at`):

Phase 1 — original conversation:
  - 14 distinct `text` rows with unique `msg_id`s, each containing exactly one Lead-side turn
  - Tool-call and lifecycle rows interspersed
  - Conversation closed by a `disconnected` status row

User prompt sent post-restart (recall-only, no tools/dispatches expected).

Phase 2 — within ~5 seconds of the post-restart prompt:
  - 8 new `text` rows with NEW `msg_id`s (e.g., `d1b39af7-276f-...`, `40d8b607-3f8a-...`, `04921406-9169-...`, plus 5 more)
  - Content matches prior Phase 1 turns but with consecutive turns concatenated together
  - 1 actual new `text` row containing the Lead's legitimate response to the recall prompt
  - Lifecycle rows interspersed

Concatenation example: a single Phase 2 row contains text from THREE separate Phase 1 turns merged into one row, with no space between the merged segments — clear evidence that the 120ms stream-chunk buffer is concatenating consecutive replay events into one persisted row.

The original Phase 1 turns were three separate Lead text messages with distinct `msg_id`s, separated by ~5-50 seconds in real time. The Phase 2 replay produced a single merged row arriving as one buffered chunk.

## Suspected root cause

Reading the source at `src/process/task/AcpAgentManager.ts` and related files:

1. ACP `session/load` is called on resume (correct — model context is restored via Claude Code's native session mechanism).

2. During `session/load`, Claude Code re-streams prior conversation events through ACP to AionUi.

3. AionUi has a `bootstrapping` flag in `AcpAgentManager.onStreamEvent` that suppresses persistence/UI events during initial connection:

       // During bootstrap (warmup), suppress UI stream events to avoid
       // triggering sidebar loading spinner before user sends a message.
       if (this.bootstrapping) {
         return;
       }

4. The flag is cleared at the start of `sendMessage`, BEFORE `initAgent` completes:

       async sendMessage(data: {...}) {
         // Allow stream events through once user actually sends a message,
         // so initAgent progress (agent_status) is visible during the wait.
         this.bootstrapping = false;
         ...
         await this.initAgent(this.options);

5. If `session/load` is still emitting replay events when `bootstrapping` flips to `false`, those events flow through `onStreamEvent` → `transformMessage` → `addOrUpdateMessage`.

6. `transformMessage` does not preserve Claude Code's original session message UUIDs — it generates new client-side UUIDs. So `addOrUpdateMessage` sees the replay events as new rows and inserts them rather than updating existing ones.

7. The 120ms stream buffering in `queueBufferedStreamTextMessage` then concatenates consecutive replay chunks that arrive close together in time, producing the content-merging artifact.

The associated code comment ("Reduce status noise: show full lifecycle only for the first turn. After first turn, only keep failure statuses to avoid reconnect chatter") indicates awareness of replay-chatter for status events, but the equivalent filtering hasn't been applied to content events.

## Suggested fix directions (in order of increasing scope)

1. **Detect-and-skip**: Extend the `bootstrapping` flag pattern. Keep replay events out of the persistence/render pipeline until the first user-initiated turn after resume actually completes (or use ACP-level signals to distinguish replay from live generation). The model context is already restored via `session/load`; the renderer doesn't need a second copy via stream events.

2. **Preserve stable IDs**: Pass Claude Code's original session message UUIDs through `transformMessage` so `addOrUpdateMessage`'s upsert semantics actually deduplicate. The IDs exist in Claude Code's session storage (`~/.claude/projects/<project>/<session-id>.jsonl`) and presumably come through in the ACP `session/load` events.

3. **Hydrate-then-stream gate**: After `session/load` completes, mark the conversation as "rehydrated" and only allow stream events for messages arriving after a fresh user input or `wake_agent` call.

Option 1 is probably the smallest change and aligns with the existing `bootstrapping` flag pattern.

## Impact

For short conversations or projects with few restarts, the bloat is bounded. For longer-running use cases — especially Team Mode conversations with persistent Leads/Teammates over multiple days — each restart adds approximately the prior conversation length in new rows. For a multi-week project with daily restarts and hundreds of turns:
- Thousands of duplicate rows in SQLite
- Multi-MB of redundant storage
- UI rendering slowdown when scrolling history
- Stored conversation no longer accurately represents the original turn structure (concatenation artifact)

This is also a blocker for using AionUi as a substrate for use cases where session continuity across restarts is core to the workflow — the visible "history replay" makes it hard to distinguish current activity from past activity at a glance.

## Scope of this report

Tested only with the Claude Code backend (ACP via `@zed-industries/claude-agent-acp` 0.23.1). The same code path (`AcpAgentManager` + `session/load`) is used by other ACP backends (Codex, Qwen, Goose), so the issue likely affects them as well, but this has not been verified.

## What's NOT affected

- Claude Code's own session storage (`~/.claude/projects/<project>/<session-id>.jsonl`) is unaffected — the model itself sees one copy of each turn via `session/load`.
- Functional behavior of the agent (model context continuity, tool use, replies) all continue working correctly after restart.
- The bug is purely in AionUi's SQLite persistence layer + UI rendering of replay events.

## Available diagnostics

The reproducer above is fully deterministic. I have a captured SQLite export of the `messages` table for the affected conversation showing both Phase 1 (originals) and Phase 2 (replay-as-new-rows) with timestamps and msg_ids. I can share an anonymized/sanitized version of it on request if it would aid diagnosis.

pmos69:
Closes #2887

## Summary

- **Problem.** When AionUi closes and reopens an existing ACP conversation, the first post-restart user message causes Claude Code's `session/load` replay events to be inserted into SQLite as new rows. The `messages` table grows by approximately the prior Lead-side turn count on every restart, and consecutive replay chunks are concatenated by the 120 ms stream buffer into a single row whose content spans multiple original turns. UI-visible symptom: prior conversation visually re-renders before the new response.

- **Verified root cause.** Two cooperating defects:
  1. `@agentclientprotocol/claude-agent-acp@0.29.2` sends `agent_message_chunk` notifications during `session/load` with `update.messageId === undefined`. `MessageTranslator.handleAgentMessageChunk` falls back to a literal `'default'` key, so every chunk in the replay batch resolves to **one** generated UUID via `resolveMsgId`. The 120 ms stream buffer in `AcpAgentManager.queueBufferedStreamTextMessage` keys by `msg_id`, so all replay chunks merge into a single buffered text message whose content is the concatenation of every prior Lead-side turn — flushed as one `addOrUpdateMessage` call with a brand-new UUID.
  2. `AcpAgentManager.bootstrapping` is meant to suppress these events, but it's a single-shot flag. `sendMessage` (line 964) unconditionally writes `false` before awaiting `initAgent`; `initAgent` re-asserts `true` only on its first invocation per process, because subsequent calls short-circuit at `if (this.bootstrap) return this.bootstrap;`. So any path that reaches `agent.start()` after the first `initAgent` resolves (e.g. the auto-reconnect from `idle`/`error` inside `AcpAgentV2.sendMessage`) runs `loadSession` with no protection.

- **Fix.** Gate `MessageTranslator.translate` while `session/load` is running. `SessionLifecycle.tryLoadOrCreate` calls `messageTranslator.setLoadingSession(true)` before the RPC and `(false)` in `finally` after — content-bearing notifications (`agent_message_chunk`, `agent_thought_chunk`, `tool_call`, `tool_call_update`, `plan`) translate to `[]` during the window. Config notifications (`current_mode_update`, `config_option_update`, `usage_update`, `available_commands_update`) are unchanged — agent state stays in sync. The scoped, protocol-layer gate replaces the fragile single-shot flag pattern and is robust to every `loadSession` call site (cold start, resume from disconnect, reconnect from `idle`/`error`/`suspended`).

- **Net diff.** ~20 LOC of production code, ~90 LOC of tests. No public API change (the new `setLoadingSession` method on `MessageTranslator` is only invoked by `SessionLifecycle`).

## Scope

- ACP (`MessageTranslator`, `SessionLifecycle`) — affects all ACP backends (Claude, Codex, Qwen, Goose, OpenCode, GitHub Copilot, Kimi, custom).
- Only Claude was verified to exhibit the `messageId=undefined` collapse; other backends route through the same code path so the gate is harmless if they don't exhibit it and protective if they do.
- No renderer / UI / DB schema changes.
- Team Mode is automatically correct — each agent (Lead, Teammates) has its own `MessageTranslator` instance, so the flag is per-session.

## Test plan

- [x] `bun run format` — clean (no changes)
- [x] `bun run lint` — 0 errors (1884 pre-existing warnings unaffected)
- [x] `bunx vitest run` — 4525 passed (was 4519, +6 new tests), 57 skipped, 3 pre-existing Windows symlink failures unrelated to ACP (`fileResolver`, `pathSafety`, `channelSendProtocol`)
- [x] 6 new `MessageTranslator` tests added covering: gate drops `agent_message_chunk` / `agent_thought_chunk` / `tool_call`; config_updates still pass through; translation resumes after `setLoadingSession(false)`; `messageMap` clears on load exit so replay-side state doesn't leak.
- [x] Manual reproducer (Windows dev build, Claude Code backend):
  - Pre-fix: pre-restart 12 rows → post-restart 17 rows (Δ +5, one row with concatenated content `"Hello, I'm LeadertwoParisblueseven…"` and a brand-new `msg_id`).
  - Post-fix: pre-restart 26 rows → post-restart 30 rows (Δ +4 — user prompt + Lead response + 2 lifecycle `agent_status` rows). No merged-content row, no new replay `msg_id`. Lead correctly recalled prior turn (`"You said: \"Say only the word 'OK'.\""`), confirming model context preserved via Claude Code's `session/load`.
  - Two restarts in a row: Δ +4 each time, cumulative growth bounded.
- [ ] Manual: regression check on a non-Claude ACP backend (Codex / Qwen / Goose / OpenCode / Copilot) — verify normal streaming still works after a fresh session start. (Not driven in this PR; same code path, no expected change since these backends weren't producing replays anyway.)
- [ ] Manual: long conversation (50+ turns) restart — sanity check that the gate window doesn't grow with replay size. (Not driven; gate is unbounded in duration but engages/disengages cleanly around the RPC.)

## Risks and limitations

- **Late-arriving replay events.** If a future SDK version sends `session/update` notifications *after* `loadSession` resolves, they'd bypass the gate. Empirically this doesn't happen with `@agentclientprotocol/claude-agent-acp@0.29.2` — the final replay chunk consistently arrived ~10 ms before the `loadSession` RPC response and no further replays followed. Worth re-checking on future SDK bumps.
- **`messageId=undefined` collapse remains latent.** If a backend ever streams two real (non-replay) live turns both with `messageId=undefined`, the 120 ms buffer would still merge them in `MessageTranslator`. Out of scope for this bug fix — flagging for a separate atomic PR.
- **No impact on already-corrupted SQLite rows from prior restarts.** Users with existing replay-duplicated rows will not see them retroactively deduplicated. A separate migration could detect and clean those (rows with identical `created_at` to ms precision and concatenated content matching a prefix of prior rows), but is out of scope here.


pmos69:
The `Unit Tests (windows-2022)` job has now been cancelled on both runs at exactly the same 10m6s mark — that's the job's `timeout-minutes: 10` (see [pr-checks.yml#L93](https://github.com/iOfficeAI/AionUi/blob/main/.github/workflows/pr-checks.yml#L93)). On the Windows runner `bun install` takes ~9 minutes from a cold cache, leaving only ~30 seconds for the actual test run before the timeout fires.

Status of the other 12 checks across both runs:

- ✅ `Code Quality` (TypeScript / Oxlint / Oxfmt)
- ✅ `Unit Tests (ubuntu-latest)` — 2m46s — runs the same vitest matrix including the 6 new tests added in this PR
- ✅ `Unit Tests (macos-14)` — 3m46s — same matrix, also clean
- ✅ `Coverage Test`
- ✅ `I18n Check`
- ✅ `Release Script Test`
- ✅ `Build Test (linux / macos-x64 / macos-arm64 / windows-x64 / windows-arm64)`
- ✅ `codecov/patch`

The actual tests that ran on the Windows job before the timeout were all `✓`. The new `MessageTranslator.test.ts` tests are non-DOM and would have run early in the vitest order — no failure was logged.

This appears to be a known infrastructure flake on this repo — looking at recent PR Checks history, most PRs show at least one `cancelled` run before a successful one on the Windows job. I've pushed an empty commit (`77327e250`) to retrigger CI on the theory that the fork's CI cache may be warmer on the third attempt.

If this keeps timing out, it may be worth a follow-up to bump `timeout-minutes` on that job — happy to open a separate PR for that if it would help.

piorpua:

pmos69:
Thanks for the heads-up @piorpua. Closing this PR — the diff is against the old monolithic layout and won't apply cleanly to the new AionUi/AionCore split.

I'll investigate whether the session-resume bug (duplicate row insertion via `session/load` replay events) still exists in the new Rust implementation in `crates/aionui-conversation/` and `crates/aionui-team/`. If it does, I'll open a fresh issue against the appropriate repo with a Rust-side reproduction and proposed fix. If it's already addressed (or no longer relevant given the rewrite), no further action needed here.

PR #2892, opened 2026-05-13
fix(acp): block file writes in plan mode

wlin111:
## Why

This came up while using Team Mode. Even when the leader session is in plan mode, the leader can still drive changes that write files. In practice that means plan mode depends too much on the agent following instructions, and some agents may still try to edit because the app continues to expose local file-write capability.

That is especially easy to hit in Team Mode because the leader is expected to be proactive and helpful. If the backend treats plan mode as guidance instead of a hard limit, AionUi can still accept `fs/write_text_file` and modify the workspace.

Plan mode should be a real local guard for agent writes, not just a prompt or backend preference. This PR adds that guard for ACP file writes.

## Summary

- Track ACP session mode locally so AionUi can enforce plan-mode write blocking at the agent execution boundary.
- Reject ACP `fs/write_text_file` requests when the local mode is `plan` and advertise `writeTextFile: false` when ACP initializes in plan mode.
- Add focused unit coverage for write rejection, mode transitions, capability advertisement, cleanup, and retry reseeding.

## Test plan

- [x] `bunx vitest run tests/unit/acpPlanModeWriteBlock.test.ts tests/unit/acpApplySessionMode.test.ts`
- [x] `bun run lint` (0 errors; existing warnings remain)
- [x] `bunx tsc --noEmit`


piorpua:

PR #2999, opened 2026-05-24
[codex] fix chat marker, WebUI QR login, and status badges

lixunqi12:
## Summary
- Treat `[[AION_FILES]]` as an attachment marker only when it appears on its own marker line in user messages.
- Preserve assistant text that merely mentions `[[AION_FILES]]` so it renders as normal markdown instead of file preview cards.
- Stabilize WebUI QR login by preferring Tailscale/reachable remote addresses consistently, routing generated QR links through the no-cache `/qr-login` static page, and accepting both camelCase and snake_case QR token payloads.
- Keep transient ACP agent status events (`connecting`, `connected`, `authenticated`, `session_active`) out of the chat transcript while preserving visible error status messages.

## Root Cause
Assistant replies can mention the literal attachment marker while explaining which screenshot was read. The renderer used a broad substring search, so it split the assistant message at that text and rendered the remaining prose as attachment previews.

For WebUI QR login, the generated remote address could be taken from the first non-internal interface, which may be a corporate/VPN 10.x address instead of Tailscale. The QR link also briefly pointed at the React hash route, which left mobile browsers vulnerable to stale SPA bundles or auth-route redirects. The generated QR URL now uses the dedicated server-rendered `/qr-login?token=...` page so token verification does not depend on the cached frontend app booting first.

For ACP status cards, `agent_status session_active` events can be persisted after the assistant's final answer and then render as the newest center message. This makes the bottom of the conversation look blocked by a green "session active" badge. These statuses are internal connection state, so the message list now filters non-error status cards out of the transcript.

## Validation
- `vitest run tests/unit/webuiQR.test.ts tests/unit/renderer/routerQrLogin.dom.test.ts tests/unit/renderer/qrLoginPage.dom.test.tsx tests/unit/chat/messageText.dom.test.tsx tests/unit/renderer/components/MessageText.dom.test.tsx tests/unit/authRoutesStatus.test.ts tests/unit/webuiServiceWorker.test.ts`
- `vitest run tests/unit/renderer/MessageList.dom.test.ts tests/unit/renderer/messageHooks.dom.test.tsx --project dom`
- `tsc --noEmit`
- `oxlint src/renderer/pages/conversation/Messages/MessageList.tsx tests/unit/renderer/MessageList.dom.test.ts`
- Built Windows package from clean worktree `clean-build-57280470` and replaced local runtime `F:\AI_tool\AionUi\resources\app.asar`; runtime hash matches the new build.
- Verified the real persisted `agent_status session_active` shape now evaluates to `shouldRender: false`, and verified the packaged `app.asar` contains the compiled status filter.


piorpua:

PR #3134, opened 2026-05-31
fix(chat): preserve visible live replies during hydration

plus61:
## Summary\n- Preserve visible in-memory live replies when DB hydration returns a hidden row with the same msg_id\n- Add a regression test for the progress/status reply disappearance case\n\n## Why\nA live assistant/team reply could render once, then disappear after message hydration if the compact DB snapshot contained a hidden text row with the same msg_id and equal content length. The text version chooser previously preferred the DB row on ties, so MessageList filtered it out as hidden.\n\n## Test plan\n- bun run format -- packages/desktop/src/common/chat/chatLib.ts tests/unit/renderer/messageMerging.dom.test.tsx\n- bunx vitest run tests/unit/renderer/messageMerging.dom.test.tsx\n- git diff --check

### DIFF
--- a/packages/desktop/src/common/chat/chatLib.ts
+++ b/packages/desktop/src/common/chat/chatLib.ts
@@ -308,6 +308,10 @@ export const mergeTextMessageContent = (
 };
 
 export const preferTextMessageVersion = (primary: IMessageText, secondary: IMessageText): IMessageText => {
+  if (primary.hidden !== secondary.hidden) {
+    return primary.hidden ? secondary : primary;
+  }
+
   const primaryIsReplace = isTextContentReplacement(primary.content);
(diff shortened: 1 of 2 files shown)


plus61:
Merge handoff: CI is green and this PR is mergeable from code/check perspective, but branch protection blocks plus61 from merging or enabling auto-merge.\n\nRequired maintainer action:\n1. Add a non-author approval review.\n2. Squash merge PR #3134.\n\nContext:\n- Fixes disappearing visible assistant/team replies after message hydration when a hidden DB row shares the same msg_id.\n- Local checks passed: format, targeted vitest, git diff --check.\n- GitHub checks all passed.\n- plus61 attempted squash merge and auto-merge; both are blocked by permissions/branch protection.

PR #3140, opened 2026-06-01
fix(conversation): map 409 already-processing to CONVERSATION_BUSY

tcp404:
## Summary

Sending a follow-up message while a turn is still streaming hit the backend's 409 \`Conversation is already processing a message\` guard, but the ACP send paths classified that as \`AIONUI_INTERNAL_ERROR\` + Retryable. Users saw a generic \"Request failed in the app\" card with a Retry tag and kept retrying — every retry tripped the same 409 ([Sentry](https://iofficeai.sentry.io/issues/?query=6a54db50581f46389faaea8f5388567c), feedback issue AIO-93).

The error code system already has a dedicated \`AIONUI_CONVERSATION_BUSY\` entry with the right copy (\"Another response is still running. Wait for it to finish before sending a new message.\") and \`resolution: wait_for_current_response\` — it just wasn't being produced from the HTTP path.

## What changed

- New \`buildSendFailureError.ts\` shared by \`AcpSendBox\` and \`useAcpInitialMessage\`, replacing two copies of the same helper.
- Recognise \`status === 409 && code === 'CONFLICT' && backendMessage.includes('already processing')\` and emit \`AIONUI_CONVERSATION_BUSY\` with \`retryable: false\` and \`resolution: wait_for_current_response\` instead of the generic \`AIONUI_INTERNAL_ERROR\` + Retryable.
- Other 409 conflicts (e.g. cancel-when-disconnected) and non-HTTP failures still fall through to the existing internal-error path.

## Test plan

- [x] \`bun run test tests/unit/renderer/buildSendFailureError.test.ts\` — new unit tests cover 409 already-processing, 409 other conflicts, 502 BAD_GATEWAY, and non-HTTP errors.
- [x] \`bun run test\` — full suite (963 passed).
- [x] \`bunx tsc --noEmit\` clean.
- [ ] Manual: send a second message while the first is still streaming → expect the busy card with the wait copy and no Retry tag.

### DIFF
--- a/packages/desktop/src/renderer/pages/conversation/platforms/acp/AcpSendBox.tsx
+++ b/packages/desktop/src/renderer/pages/conversation/platforms/acp/AcpSendBox.tsx
@@ -1,7 +1,6 @@
 import { ipcBridge } from '@/common';
 import type { IConversationMcpStatus } from '@/common/config/storage';
 import { isBackendHttpError } from '@/common/adapter/httpBridge';
-import type { AgentStreamErrorInfo } from '@/common/chat/chatLib';
 import { isSideQuestionSupported } from '@/common/chat/sideQuestion';
 import { parseError, uuid } from '@/common/utils';
 import AgentModeSelector from '@/renderer/components/agent/AgentModeSelector';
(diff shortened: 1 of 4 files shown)


tcp404:
迁移到符合规范的分支 boii/fix-aio-93，将由新 PR 替代。

Issue #3075, opened 2026-05-27
[Bug]: STT config entered in Capabilities → Speech to Text is not persisted (lost on restart, mic icon never appears)


szafranski:
### Platform

macOS (Apple Silicon)

### AionUi Version

2.1.5

### Bug Description

**Environment**
- AionUi 2.1.5 (desktop)
- macOS 26.5, Apple Silicon (arm64)

### Steps to Reproduce

**Steps to reproduce**
1. Settings → Capabilities → Tools → Speech to Text.
2. Toggle STT on, fill Provider (OpenAI Whisper), API Key, Base URL, Model, Default Language.
3. Restart AionUi.

### Expected Behavior

**Expected**
The GUI save should persist `tools.speechToText` so the mic icon shows and settings survive a restart.

### Actual Behavior

**Actual**
- The settings are gone after restart.
- The microphone icon never appears in the chat composer.
- The key `tools.speechToText` is **never written** to the config store
  (`~/Library/Application Support/AionUi/config/aionui-config.txt`). After configuring + restarting, that file contains no `tools.speechToText` entry at all (no provider, base URL, or model anywhere on disk).

### Additional Context

*Confirmation that the read path works but the save path doesn't**
Manually injecting a `tools.speechToText` object into `aionui-config.txt`:
- makes the mic icon appear in the chat composer, and
- survives restarts.
So the app reads the key correctly; the GUI just doesn't write it. The save handler (`G.set("tools.speechToText", …)`) appears to fail silently (its `.catch` only `console.error`s).

**Note:** The mic icon only renders inside an active conversation, not on the welcome/home screen — minor, but worth documenting.

szafranski:
## Update — bug is deeper than "GUI doesn't persist to file"

Further investigation shows that even when the STT config is *manually* written into `aionui-config.txt` (so the renderer correctly reads it on next launch, the mic icon appears, etc.), the **backend (`aioncore`) still reports `STT_DISABLED`** when the request reaches it. So persistence is only half the story — there's also no config propagation between the renderer's settings store and `aioncore`.

### Reproduction

1. Manually inject a valid `tools.speechToText` object into `~/Library/Application Support/AionUi/config/aionui-config.txt` (matching the default shape from `CapabilitiesSettings`), with `enabled: true`, a working provider, model, API key and base URL.
2. Restart AionUi. The renderer reads the config correctly (mic icon appears in the chat composer, settings page shows the values).
[...]

PR #3159, opened 2026-06-02
fix(settings): surface STT config save failures instead of swallowing them

ktwu01:
## What

The Speech-to-Text save handler in `ToolsModalContent.tsx` caught persistence errors with a bare `console.error`, giving the user **no visible feedback**. The form updates optimistically, the user assumes the setting was saved, and only discovers it was lost on the next restart — the mic icon never appears.

## Changes

`packages/desktop/src/renderer/components/settings/SettingsModal/contents/ToolsModalContent.tsx`:

- **Surface save failures**: show an error toast via `mcpMessage.error(...)` when `configService.set('tools.speechToText', …)` rejects, reusing the `Message` instance already created and rendered in this component (same idiom used elsewhere in the file for image-generation errors).
- **Gate the change event on success**: move the `SPEECH_TO_TEXT_CONFIG_CHANGED_EVENT` dispatch into the `.then()` success path. The listener in `SpeechInputButton.tsx` re-reads the persisted config (`configService.get('tools.speechToText')`) to decide whether to show the mic icon, so it should only react once the config is actually persisted — never to a config that failed to save.

No new i18n keys are introduced; the message falls back to the existing `common.saveFailed` key with an inline `defaultValue`.

## Scope

This addresses the **renderer half** of #3075 (the silently-swallowed save failure). The reporter also describes a backend half — `aioncore` not observing the persisted STT config and returning `STT_DISABLED`. That config-propagation work lives in the bundled backend and is out of scope for this PR.

## Testing

Manual: trigger a failing `configService.set` and confirm an error toast now appears instead of a silent console log. The change is isolated to the save handler's promise chain.

Closes part of #3075

### DIFF
--- a/packages/desktop/src/renderer/components/settings/SettingsModal/contents/ToolsModalContent.tsx
+++ b/packages/desktop/src/renderer/components/settings/SettingsModal/contents/ToolsModalContent.tsx
@@ -525,18 +525,33 @@ const ToolsModalContent: React.FC = () => {
     void loadConfigs();
   }, []);
 
-  const updateSpeechToTextConfig = useCallback((updater: (current: SpeechToTextConfig) => SpeechToTextConfig) => {
-    setSpeechToTextConfig((current) => {
-      const next = normalizeSpeechToTextConfig(updater(current));
-      configService.set('tools.speechToText', next).catch((error) => {
-        console.error('Failed to save speech-to-text config:', error);
+  const updateSpeechToTextConfig = useCallback(
+    (updater: (current: SpeechToTextConfig) => SpeechToTextConfig) => {
+      setSpeechToTextConfig((current) => {
+        const next = normalizeSpeechToTextConfig(updater(current));
+        configService
+          .set('tools.speechToText', next)
+          .then(() => {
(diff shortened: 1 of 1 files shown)


lornestack:
Thank you for the contribution, and for the careful analysis in the PR description — the renderer/backend split you identified was accurate.

Closing this PR because the underlying issue (#3075) has since been resolved by other changes, and this branch now conflicts with `main`:

- **Persistence half**: config storage was migrated from the local `aionui-config.txt` file to the backend `client_preferences` store. `configService.set('tools.speechToText', ...)` now persists through the backend, and we've verified the config survives restarts.
- **Backend half** (`STT_DISABLED` despite a saved config): root-caused to a preference key mismatch — the backend read `speechToText` while the UI saves `tools.speechToText` — plus a request-format mismatch on `/api/stt`. Fixed in iOfficeAI/AionCore#400, iOfficeAI/AionCore#448, and #3274.

The one piece of this PR that is still relevant on its own merits — surfacing save failures with a toast instead of a silent `console.error` — is a small UX improvement we'd still welcome. If you'd like to rebase that part onto current `main` as a standalone polish PR, we'd be happy to review it.

PR #3179, opened 2026-06-03
fix(preview): normalize preview navigation URLs

CottChen:
## Summary

Fix three preview/WebUI regressions:

1. Office preview proxy root URLs returning 404 in WebUI/browser mode after preview startup succeeds.
2. HTML preview hash anchors such as `#s2` navigating the host app to the domain/root path instead of scrolling inside the preview iframe.
3. Direct WebUI access can inherit `X-Frame-Options: DENY` from Office/PPT preview proxy responses, blocking iframe-based previews.

## Problem 1: Office Preview Proxy 404

User logs showed the preview start endpoints returned HTTP 200, but the iframe then loaded proxy root URLs with a trailing slash and received 404 from aioncore:

- `POST /api/excel-preview/start` -> `200`
- `GET /api/office-watch-proxy/50753/` -> `404`
- `POST /api/ppt-preview/start` -> `200`
- `GET /api/ppt-proxy/50918/` -> `404`

The issue affected both Excel and PPT preview. This looked like a proxy route/root URL handling regression rather than corrupted Office files or a frontend render-only issue.

### Root Cause

In WebUI/browser mode, `OfficeWatchViewer` normalized proxy root URLs to include a trailing slash:

- `/api/office-watch-proxy/{port}/`
- `/api/ppt-proxy/{port}/`

aioncore currently has explicit root routes without the trailing slash:

- `/api/office-watch-proxy/{port}`
- `/api/ppt-proxy/{port}`

The trailing-slash form was missed at the backend router layer and returned 404 before the proxy handler could run.

### Fix

- Keep WebUI proxy root URLs slashless in `OfficeWatchViewer` so iframe `src` matches the backend root route.
- Preserve non-root proxy suffixes such as `/index.html`.
- Keep Electron desktop behavior unchanged: Electron still opens direct `http://127.0.0.1:{port}/` URLs.
- Add a WebHost reverse-proxy fallback that strips trailing slashes only for Office/PPT proxy root paths before forwarding to aioncore.
- Add regression coverage for both URL normalization and WebHost proxy forwarding.

## Problem 3: Direct WebUI Office Preview Blocked by X-Frame-Options

When accessing WebUI directly, for example `http://<host>:25808/`, Office/PPT preview responses proxied through AionUi can include:

```http
x-frame-options: DENY
```

WebUI loads Office/PPT preview pages in iframes. `DENY` blocks iframe rendering even when the preview service itself started successfully.

### Root Cause

In direct WebUI deployments, there may be no external reverse proxy layer available to adjust response headers. AionUi preserved the backend preview response headers unchanged, so iframe-blocking `X-Frame-Options` values were forwarded to the browser.

### Fix

- Add a scoped `officeProxyFrameOptions` option for Office/PPT preview proxy responses only.
- Keep the default behavior as `preserve`, so existing deployments continue forwarding backend headers unchanged.
- Allow direct WebUI deployments to set `sameorigin`, `deny`, or `remove`.
- Wire the option through desktop WebUI startup, `@aionui/web-host`, `@aionui/web-cli`, and `scripts/webui.ts`.

### Reverse Proxy Guidance

If AionUi WebUI is served behind a reverse proxy such as Caddy or Nginx, prefer configuring the preview response headers in the reverse proxy layer. For example, Caddy can strip or override `X-Frame-Options` only for:

- `/api/office-watch-proxy/*`
- `/api/ppt-proxy/*`

In that setup, no AionUi-side configuration is needed.

The new AionUi option is mainly for deployments that do not use a reverse proxy and access WebUI directly, for example `http://<host>:25808/`.

Example direct WebUI usage:

```bash
AIONUI_OFFICE_PROXY_FRAME_OPTIONS=sameorigin aionui-web start --allow-remote
```

Equivalent CLI flag:

```bash
aionui-web start --allow-remote --office-proxy-frame-options sameorigin
```

## Problem 2: HTML Preview Hash Anchor Navigation

When previewing HTML, clicking an in-document anchor such as:

```html
<a href="#s2">Section 2</a>
```

could navigate the host app/browser to the domain/root path hash instead of scrolling inside the HTML preview iframe.

### Root Cause

The HTML preview uses iframe/data-URL rendering paths. Same-document hash links can be resolved against the host/base URL rather than staying scoped to the preview document, causing the outer app route/hash to change.

### Fix

- Inject a small hash-anchor guard into HTML preview content.
- Only intercept same-document hash links (`href="#..."`).
- Scroll to matching `id` or legacy `name` targets inside the preview document.
- Do not rewrite the host window location.
- Do not intercept external, absolute, relative file, mailto, or non-hash links.
- Cover both browser `srcDoc` and Electron data-URL HTML rendering paths.

## Validation

GitHub Actions on this PR passed:

- Code Quality
- Unit Tests (Ubuntu/macOS/Windows)
- Coverage Test
- I18n Check
- Build Test (macos-arm64)
- Build Test (macos-x64)
- Build Test (windows-x64)
- Build Test (windows-arm64)
- Build Test (linux)
- Release Script Test
- codecov/patch

Manual/package validation:

- Verified the Office preview fix with a generated Windows x64 installer.
- Verified the HTML hash-anchor behavior manually.
- Added WebHost coverage for preserving, overriding, and removing Office/PPT proxy `X-Frame-Options` headers.
- Fork manual Windows x64 build succeeded:
  https://github.com/CottChen/AionUi/actions/runs/26880315448

## Platform Impact

- WebUI/browser mode: fixes Windows/Linux/macOS because the affected code is shared renderer/WebHost code.
- Direct WebUI deployments can opt into `AIONUI_OFFICE_PROXY_FRAME_OPTIONS=sameorigin` or `remove` when iframe preview is blocked by forwarded `X-Frame-Options`.
- Reverse-proxied deployments should normally configure these headers in Caddy/Nginx instead of changing AionUi runtime config.
- Electron desktop mode: Office proxy behavior is unchanged and remains direct localhost. HTML hash-anchor guard also applies, but only to same-document hash links inside HTML preview.


CottChen:
Closing this PR because the related changes are no longer intended to be proposed from origin/main. We will keep the fork main aligned with upstream/main and use a dedicated development branch instead.

PR #3186, opened 2026-06-04
fix(webui): allow office proxy frame options

CottChen:
## Summary

- Add a configurable `officeProxyFrameOptions` mode for Office/PPT preview proxy responses in WebUI mode.
- Preserve the existing backend `X-Frame-Options` behavior by default, while allowing `sameorigin`, `deny`, or `remove` when needed.
- Wire the option through desktop WebUI startup, `@aionui/web-host`, `@aionui/web-cli`, and `scripts/webui.ts`.

## Reverse proxy note

If AionUi WebUI is served behind a reverse proxy such as Caddy, prefer configuring or overriding the frame-related response headers in the reverse proxy layer. For example, Caddy can strip or set `X-Frame-Options` for `/api/office-watch-proxy/*` and `/api/ppt-proxy/*` while leaving the rest of the site unchanged.

This new AionUi-side option is mainly for deployments that access WebUI directly, for example `http://<host>:25808/`, where there is no reverse proxy layer available to adjust the preview proxy response headers.

Example direct WebUI usage:

```bash
AIONUI_OFFICE_PROXY_FRAME_OPTIONS=sameorigin aionui-web start --allow-remote
```

Equivalent CLI flag:

```bash
aionui-web start --allow-remote --office-proxy-frame-options sameorigin
```

## Test plan

- [x] `bun run format`
- [x] `bun run lint` (passes with existing warnings)
- [x] `bunx tsc --noEmit`
- [ ] `bunx vitest run --config packages/web-host/vitest.config.ts packages/web-host/src/static-server.unit.test.ts` (the package-level existing test file currently times out in existing hook setup before reaching the new assertions locally)

### DIFF
--- a/packages/desktop/src/process/utils/webuiConfig.ts
+++ b/packages/desktop/src/process/utils/webuiConfig.ts
@@ -10,7 +10,7 @@ import * as path from 'path';
 import { networkInterfaces } from 'os';
 import { getSystemDir } from './initStorage';
 import { httpRequest } from '@/common/adapter/httpBridge';
-import { startWebHost, type WebHostHandle } from '@aionui/web-host';
+import { normalizeOfficeProxyFrameOptions, startWebHost, type WebHostHandle } from '@aionui/web-host';
 import { getDataPath } from './utils';
 
 const WEBUI_CONFIG_FILE = 'webui.config.json';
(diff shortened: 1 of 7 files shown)


CottChen:
Closing this duplicate PR because the same WebUI frame-options change is already included in #3179. I will keep the deployment guidance there instead.

PR #3279, opened 2026-06-10
fix(acp): bump bundled codex-acp to 0.15.0

### LINKED ISSUE #3225: [Bug]: Codex agent fails to start (USER_AGENT_STARTUP_FAILED) — bundled codex-acp 0.14.0 rejects service_tier = "default" (fixed in codex-acp 0.15.0)
### Platform

macOS (Apple Silicon)

> Note: the root cause is **platform-agnostic** (config parsing, not a native-binary issue). Reproduced on Apple Silicon, but it applies to any platform where Codex has written `service_tier = "default"`.

### AionUi Version

2.1.12

### Bug Description

Starting the Codex agent fails immediately with **"The selected agent failed to start"** / `USER_AGENT_STARTUP_FAILED`. The `codex-acp` process exits with code 1 **before** the ACP `initialize` handshake completes, so any message to the Codex agent fails.

**Root cause:** AionUi pins `@zed-industries/codex-acp@0.14.0`, whose vendored codex-core parses the `service_tier` key in `~/.codex/config.toml` **strictly** as an enum (`fast` | `flex`) and rejects the value `"default"`. However, current Codex CLI / Codex.app write `service_tier = "default"` as a valid sentinel — upstream codex treats this field as `Option<String>`, where `"default"` means "standard routing / no tier". Because `~/.codex/config.toml` is **shared** between Codex CLI/Desktop and AionUi's bundled codex-acp adapter, a value written by Codex makes AionUi's adapter unparseable.

This is a **config-schema drift** regression: it starts happening as soon as a recent Codex writes `service_tier = "default"`, even though AionUi itself did not change. **It is already fixed upstream in `codex-acp` 0.15.0** (verified below).

### Steps to Reproduce

1. Have a recent Codex CLI / Codex.app that writes `service_tier = "default"` into `~/.codex/config.toml` (or add that line manually).
2. In AionUi, start/select the Codex agent (or send it any message).
3. The agent fails to start.

Minimal standalone reproduction (isolates the exact failing component — no AionUi needed):

```bash
# Reproduce on the version AionUi bundles (0.14.0):
H=$(mktemp -d); printf 'service_tier = "default"\n' > "$H/config.toml"
echo '{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{}}}' \
  | CODEX_HOME="$H" npx -y @zed-industries/codex-acp@0.14.0
# -> Error: error loading config: .../config.toml:1:16: unknown variant `default`, expected `fast` or `flex`
#    process exits with code 1

# Same input on 0.15.0 succeeds:
echo '{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{}}}' \
  | CODEX_HOME="$H" npx -y @zed-industries/codex-acp@0.15.0
# -> {"jsonrpc":"2.0","result":{"protocolVersion":1,...,"agentInfo":{"name":"codex-acp",...,"version":"0.15.0"}},"id":0}
```

### Expected Behavior

The Codex agent starts normally regardless of whether the shared `~/.codex/config.toml` contains `service_tier = "default"` — a value that the official Codex CLI/Desktop write and accept.

### Actual Behavior

`codex-acp` 0.14.0 exits with code 1 before the ACP handshake:

```
Error: error loading config: .../config.toml:<line>:16: unknown variant `default`, expected `fast` or `flex`
```

AionUi surfaces this as "The selected agent failed to start" / `USER_AGENT_STARTUP_FAILED` (agent process exited before the initialize handshake completed, exit code 1).

### Additional Context

- **Suggested fix:** bump the pinned codex-acp version. The default is set in `scripts/prepare-managed-acp-tools.sh`:
  ```
  CODEX_ACP_VERSION="${CODEX_ACP_VERSION:-0.14.0}"
  ```
  Bumping it to `0.15.0` (or later) resolves the issue. Note that the in-app migration also rewrites the Codex agent args to `@zed-industries/codex-acp@0.14.0`, so that pinned string needs bumping as well.
- **Verified** on macOS (Apple Silicon): `0.14.0` rejects `service_tier = "default"`; `0.15.0` accepts it and completes `initialize` (returns `agentInfo.version: 0.15.0`).
- **User-side workaround** until the pin is bumped: set `service_tier = "fast"` (or `"flex"`) in `~/.codex/config.toml`. Downside: this also changes the tier used by Codex CLI/Desktop, so bumping the bundled adapter is the cleaner fix.
- This is a **different root cause** than #1592 (a missing `@zed-industries/codex-acp-darwin-x64` optional dependency / `ERR_MODULE_NOT_FOUND` on Intel), despite the similar "exited during startup (code: 1)" symptom string.

szafranski:
## Description

Bumps the default `CODEX_ACP_VERSION` in `scripts/prepare-managed-acp-tools.sh` from `0.14.0` to `0.15.0`, and updates the version-named fixture directories in `tests/unit/assets/verifyBundledAioncoreResources.test.ts` to match.

**Why:** codex-acp 0.14.0 fails to start with `USER_AGENT_STARTUP_FAILED` when the user's shared `~/.codex/config.toml` contains `service_tier = "default"` — its config parser uses a strict enum that only accepts `"fast"`/`"flex"`, so any other value (including the one the official Codex CLI itself writes) aborts startup. codex-acp 0.15.0 relaxed the field and accepts any value (zed-industries/codex-acp v0.15.0 release). Full diagnosis with stderr logs is in #3225.

I verified the fix locally on macOS (arm64) by swapping the managed runtime tool under `~/.aionui/runtime/managed-tools/acp/codex-acp/` to the 0.15.0 npm artifacts — the Codex agent then starts and works normally with the same `~/.codex/config.toml`.

## Related Issues

- Closes #3225

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux
- [x] My code follows the project's code style guidelines
- [x] My changes generate no new warnings or errors

Local checks run: `bun run format` (no diff), `bun run lint` (0 errors), `tsc --noEmit` (clean), `vitest run tests/unit/assets/verifyBundledAioncoreResources.test.ts` (6/6 passed).

## Additional Context

- The bundled binaries published to `static.aionui.com/managed/acp` will need to be re-published by maintainers for the new version to ship — this PR changes the source default so the existing pipeline picks up 0.15.0.
- The builtin agent definitions in AionCore's DB migrations (`crates/aionui-db/migrations/008_builtin_acp_agents_use_npx.sql`) also pin `@zed-industries/codex-acp@0.14.0` for the `npx` fallback path. If desired, I'm happy to submit a companion PR to iOfficeAI/AionCore adding a new migration that bumps those args to 0.15.0.

### DIFF
--- a/scripts/prepare-managed-acp-tools.sh
+++ b/scripts/prepare-managed-acp-tools.sh
@@ -18,7 +18,7 @@ Environment variables:
   MANAGED_ACP_WRITE_ROOT_MANIFEST
                                Optional. true/false. Default: false
   MANAGED_ACP_NPM_VERSION      Optional. Exact npm version expected in PATH.
-  CODEX_ACP_VERSION            Optional. Default: 0.14.0
+  CODEX_ACP_VERSION            Optional. Default: 0.15.0
   CLAUDE_ACP_VERSION           Optional. Default: 0.39.0
 EOF
 }
@@ -59,7 +59,7 @@ MANAGED_ACP_TARGETS="${MANAGED_ACP_TARGETS:-darwin-arm64,darwin-x64,linux-x64,li
 MANAGED_ACP_OVERWRITE="${MANAGED_ACP_OVERWRITE:-false}"
 MANAGED_ACP_WRITE_ROOT_MANIFEST="${MANAGED_ACP_WRITE_ROOT_MANIFEST:-false}"
 MANAGED_ACP_NPM_VERSION="${MANAGED_ACP_NPM_VERSION:-}"
-CODEX_ACP_VERSION="${CODEX_ACP_VERSION:-0.14.0}"
+CODEX_ACP_VERSION="${CODEX_ACP_VERSION:-0.15.0}"
 CLAUDE_ACP_VERSION="${CLAUDE_ACP_VERSION:-0.39.0}"
 
 sanitize_version() {
--- a/tests/unit/assets/verifyBundledAioncoreResources.test.ts
+++ b/tests/unit/assets/verifyBundledAioncoreResources.test.ts
@@ -26,7 +26,7 @@ describe('verifyBundledAioncoreResources', () => {
     mkdirSync(nodeRoot, { recursive: true });
     writeFileSync(join(nodeRoot, 'node.exe'), '', { flush: true });
 
-    codexRoot = join(managedResourcesDir, 'acp', 'codex-acp', '0.14.0', 'win32-x64');
+    codexRoot = join(managedResourcesDir, 'acp', 'codex-acp', '0.15.0', 'win32-x64');
     mkdirSync(codexRoot, { recursive: true });
     writeFileSync(join(codexRoot, 'manifest.json'), JSON.stringify({ entrypoint: 'codex-acp.exe', path_entries: [] }), {
       flush: true,
@@ -84,7 +84,7 @@ describe('verifyBundledAioncoreResources', () => {
       flush: true,
     });
 
-    const darwinCodexRoot = join(darwinManagedResourcesDir, 'acp', 'codex-acp', '0.14.0', 'darwin-arm64');
+    const darwinCodexRoot = join(darwinManagedResourcesDir, 'acp', 'codex-acp', '0.15.0', 'darwin-arm64');
     mkdirSync(darwinCodexRoot, { recursive: true });
     writeFileSync(join(darwinCodexRoot, 'manifest.json'), JSON.stringify({ entrypoint: 'codex-acp' }), {
       flush: true,
@@ -150,7 +150,7 @@ describe('verifyBundledAioncoreResources', () => {
     });
 
     expect(result.missing).toContain(
-      'bundled-aioncore/win32-x64/managed-resources/acp/codex-acp/0.14.0/win32-x64/codex-acp.exe'
+      'bundled-aioncore/win32-x64/managed-resources/acp/codex-acp/0.15.0/win32-x64/codex-acp.exe'
     );
   });
 });


lornestack:
Thanks for the detailed diagnosis here — the root-cause writeup in #3225 (strict `service_tier` enum in codex-acp 0.14.0 vs. the `"default"` value written by official Codex) is spot-on and was really useful.

One thing worth flagging before this gets merged: I don't think the file changed in this PR is actually in the desktop packaging path, and the underlying issue already appears to be fixed through a different route.

**`scripts/prepare-managed-acp-tools.sh` isn't used to bundle codex-acp into the app.** It's a standalone S3/CDN publish script (not referenced anywhere in `.github/` or the package scripts). The desktop build bundles managed resources via `packages/shared-scripts/src/prepare-aioncore.js`, which invokes the aioncore binary's own `prepare-managed-resources` subcommand. So the codex-acp version that actually ships is pinned **inside the aioncore binary**, controlled by `aioncoreVersion` in `package.json` — not by this script's default.

**The fix already landed via the aioncore bump.** `main` now pins `aioncoreVersion: v0.1.33`, and that release bundles codex-acp **0.16.0** (see `ManagedAcpToolId::CodexAcp.version()` in `crates/aionui-runtime/src/acp_tool_runtime/types.rs`, which moved 0.14.0 → 0.16.0 between v0.1.30 and v0.1.33). 0.16.0 includes the relaxed `service_tier` field, so the original `USER_AGENT_STARTUP_FAILED` should be resolved.

The "in-app migration rewrites the agent args to `@…/codex-acp@0.14.0`" concern from the issue is also handled: AionCore migration `009_builtin_managed_acp_tools` clears the builtin Codex agent `args` to `[]` and resolves the command through the managed tool (now 0.16.0), superseding the hardcoded npx pin in migration 008.

Given that, this PR as-is wouldn't change what ships, and the underlying issue is already resolved by the aioncore v0.1.33 bump — so I'd suggest closing it. Either way, huge thanks for the investigation — it made tracking this down much easier.

szafranski:
Happy to help, closing then

Issue #3363, opened 2026-06-17
bug: team_spawn_agent fails with 'missing field name' for non-ACP CLI agents (codex/claude/gemini)


muzimu217:
## Description

When creating a team agent via `team_spawn_agent` with `agent_type` set to `codex`, `claude`, `qoder`, `gemini`, or any non-ACP/aionrs CLI backend, the operation fails with:

```
MCP tool error: JSON-RPC error -32602: failed to deserialize parameters: missing field `name`
```

## Steps to Reproduce

1. Launch AionUi with team feature enabled
2. Attempt to create a teammate:
   - Via tool call: `team_spawn_agent(name="MyAgent", agent_type="codex")`
   - Or via UI: select any non-ACP CLI agent (e.g., Codex CLI, Claude Code, Qoder, Gemini)
3. The request fails with deserialization error

## Expected Behavior

The agent should be created successfully with the specified backend type.

## Actual Behavior

- MCP server returns `-32602` (Invalid params): `"missing field name"`
- The `name` field IS provided in the request, but the error still occurs
- Only works when `agent_type` is omitted (defaults to `aionrs`)

## Environment

- OS: macOS
- AionUi Version: Latest (Electron app)
- Affected agent types: `codex`, `claude`, `qoder`, `gemini` (all non-aionrs/non-acp CLI agents)

## Root Cause

**File**: `out/renderer/assets/jobAgentMeta-DGlWpxnc.js`  
**Function**: `Ye()` — agent config builder

The function has three conditional branches when processing CLI agents:

```javascript
if (g === "aionrs") {
  // ✅ Works — builds agent_config with name field
  y = { backend: E.id, name: E.name || u?.name || "Aion CLI", ... };
} else if (u?.agent_type === "acp") {
  // ✅ Works — builds agent_config with name field
  y = { backend: g, name: u.name || N, ... };
} else {
  // ❌ BUG: y (agent_config) is NEVER assigned
  u && (C = Q(g));  // Only updates resolvedAgentType, no agent_config created
}
```

When the agent type is `codex`, `claude`, `qoder`, `gemini`, or any non-aionrs/non-acp type:
1. The `else` branch executes
2. Only `C` (resolvedAgentType) is updated
3. `y` (agent_config) remains `undefined`
4. Serialized request is missing the `name` field
5. MCP server deserialization fails

## Proposed Fix

```javascript
// Before (broken)
} else {
  u && (C = Q(g));
}

// After (fixed)
} else {
[...]

PR #3364, opened 2026-06-17
fix: fail-fast on legacy agent_type in resolveCronAgentConfig

muzimu217:
## Summary

Addresses the cron agent config path for Issue #3363.

**Important**: This PR fixes the `resolveCronAgentConfig` helper to properly handle legacy agent types, but **this is NOT the root cause of Issue #3363**. See [Root Cause Analysis](#root-cause-analysis) below.

## Changes

1. **Fail-fast on legacy agent_type**: When `agent_type` is not `"acp"` or `"aionrs"` (e.g., legacy `"codex"`), throw a clear error instead of silently normalizing
2. **Add unit tests**: Cover canonical ACP format `{ agent_type: "acp", backend: "codex" }` and legacy row rejection behavior

### Verified Facts

| Test | Result |
|------|--------|
| `team_spawn_agent(agent_type="codex", name="Test")` | ✅ **Success** — `agent_type: "codex"` is allowed |
| `team_spawn_agent(agent_type="codex")` (no name) | ❌ `missing field name` |
| `team_spawn_agent()` (empty) | ❌ `missing field name` |

### Conclusion

The error `"missing field name"` is caused by **JSON-RPC arguments missing the required `name` field**, NOT by `agent_type: "codex"` being rejected.

The backend evidence confirms this:
- `SpawnAgentInput.name` is **required** ([tools.rs:60](https://github.com/iOfficeAI/AionCore/blob/main/crates/aionui-team/src/mcp/tools.rs#L60))
- `agent_type` / `backend` are **optional** and merged ([server.rs:617](https://github.com/iOfficeAI/AionCore/blob/main/crates/aionui-team/src/mcp/server.rs#L617))
- `codex` in team spawn collapses to ACP ([spawn_support.rs:12](https://github.com/iOfficeAI/AionCore/blob/main/crates/aionui-team/src/service/spawn_support.rs#L12))

### Real Root Cause Location

The real issue is in the **tool-call generation or parameter passing chain**:

```
Leader agent calls team_spawn_agent
→ MCP tool schema (name is required)
→ Tool-call generation (does LLM correctly generate name parameter?)
→ JSON-RPC serialization
→ MCP server deserialization
→ SpawnAgentInput.name is missing
```

**Next step**: Capture the complete JSON-RPC `tools/call` payload when the failure occurs to determine if:
1. The LLM is not generating the `name` parameter, OR
2. A bridge/adapter is stripping/corrupting the parameters

## What This PR Does

- ✅ Fixes `resolveCronAgentConfig` to properly reject legacy agent types (per reviewer feedback)
- ✅ Adds unit tests for canonical and legacy formats
- ❌ Does NOT fix the root cause of Issue #3363 (which is in the MCP tool-call generation path)

## Testing

1. Unit tests verify canonical ACP format works
2. Unit tests verify legacy format is rejected with clear error
3. All CI checks pass

### DIFF
--- a/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/resolveCronAgentConfig.ts
+++ b/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/resolveCronAgentConfig.ts
@@ -84,6 +84,13 @@ export function resolveCronAgentConfig(input: ResolveCronAgentConfigInput): Reso
       };
     } else if (agent) {
       resolvedAgentType = resolveSupportedConversationType(backend);
+      agent_config = {
+        backend,
+        name: agent.name || (backend.charAt(0).toUpperCase() + backend.slice(1)),
+        mode: getMode(backend),
+        model_id,
+        workspace,
+      };
     }
   } else if (agentKind === 'preset') {
     const assistant = presetAssistants.find((item) => item.id === agentId);
--- a/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/resolveCronAgentConfig.ts
+++ b/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/resolveCronAgentConfig.ts
@@ -7,7 +7,10 @@
 import type { ICreateCronJobParams, ICronAgentConfig } from '@/common/adapter/ipcBridge';
 import type { Assistant } from '@/common/types/agent/assistantTypes';
 import type { AgentMetadata } from '@renderer/utils/model/agentTypes';
-import { normalizeSupportedAgentSelection, resolveSupportedConversationType } from '@renderer/utils/model/agentTypeSupportPolicy';
+import {
+  normalizeSupportedAgentSelection,
+  resolveSupportedConversationType,
+} from '@renderer/utils/model/agentTypeSupportPolicy';
 
 type SelectedAionrsProvider = {
   id?: string;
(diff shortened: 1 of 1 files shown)


piorpua:
I don’t think this should be fixed by normalizing legacy runtime types inside `resolveCronAgentConfig`.

Current architecture only supports `agent_type = "acp"` and `agent_type = "aionrs"` for new conversations/jobs. Vendor CLIs such as Codex, Claude, Gemini, etc. should be represented as:

```json
{ "agent_type": "acp", "backend": "codex" }
```

not as:

```json
{ "agent_type": "codex" }
```

Legacy runtime types are kept only for historical/read-only conversations. The normal cron selector path already uses `useConversationAgents()`, which consumes `/api/agents`; the backend `AgentService::list_agents()` filters to `supports_new_conversation()`, and the frontend filters again via `isSupportedNewConversationAgent()`. So `resolveCronAgentConfig()` should normally only receive canonical `acp`/`aionrs` agent metadata.

If `agent_type = "codex"` reaches this helper, that suggests an upstream data-shape violation or a caller bypassing the normal selector path. Silently converting that legacy shape here would preserve an unsupported input format and may hide the actual source of the bad metadata.

Main already handles the canonical ACP shape correctly. I’d suggest not merging this normalization as-is. If we want extra defense here, it should fail fast or ignore unsupported legacy runtime types, with a regression test proving that behavior. If the original issue is still reproducible, we need the actual request payload/source path that sends the legacy shape.

piorpua:
One more point on existing/legacy cron data: if a persisted cron job still has an old runtime shape like `agent_type: "codex"`, I think the safer behavior is to treat that as legacy data that needs to be recreated or migrated into the current contract, not silently normalized during submit.

For display/history purposes we can tolerate legacy rows, but when editing/saving/running a cron job it should either require the user to select a currently supported agent again, or explicitly migrate the stored metadata to the canonical shape (`agent_type: "acp"` + `agent_config.backend: "codex"`). That keeps the boundary clear: legacy runtime rows remain historical/read-only, while new executable cron jobs use the current runtime contract.

PR #3378, opened 2026-06-20
fix(conversation): keep assistant streaming auto-scroll pinned

### LINKED ISSUE #3377: Assistant response auto-scroll can intermittently stop while streaming
## Summary

Assistant response auto-scroll can intermittently stop following the bottom while a streamed reply is growing, even when the user has not intentionally scrolled away.

## Observed behavior

During longer assistant responses, the message list can sometimes remain slightly above the newest content instead of staying pinned to the bottom. The issue is intermittent because normal content resize events often keep auto-follow working.

## Expected behavior

If the user is already at or near the bottom when an assistant response starts streaming, the message list should continue following the bottom as the assistant message updates. If the user manually scrolls upward, auto-follow should stop until they return to the bottom or use the scroll-to-bottom control.

## Code notes

The current auto-scroll hook has an explicit bottom-sync path for appended user messages (`position === 'right'`). Assistant streaming usually updates the latest left-side text message with the same `msg_id`, so the message count often does not increase. In that case, auto-follow relies mostly on resize observation.

That means the UI can behave correctly most of the time, but still has a missing fallback when a streamed assistant update is applied and a resize/follow event is missed or arrives at the wrong moment.

Relevant files:

- `packages/desktop/src/renderer/pages/conversation/Messages/useAutoScroll.ts`
- `packages/desktop/src/renderer/pages/conversation/Messages/hooks.ts`
- `tests/unit/renderer/useAutoScroll.dom.test.tsx`
- `tests/unit/renderer/messageListStreaming.dom.test.tsx`

## Suggested fix

Keep auto-follow active when the latest assistant message object changes, as long as the user has not manually scrolled away from the bottom. Preserve the existing behavior where manual scrolling disables auto-follow.

Add a regression test that updates the latest left-side assistant message without relying on `ResizeObserver`, and assert that the scroller remains pinned to the bottom.

szafranski:
## Summary

- keep assistant streaming auto-follow active when the latest message is updated
- preserve the existing manual-scroll escape hatch
- add a regression test for assistant updates when `ResizeObserver` does not fire

Fixes #3377.

## Testing

- `bun run format`
- `bun run lint` (0 errors; existing warnings)
- `bunx tsc --noEmit`
- `bun run i18n:types`
- `node scripts/check-i18n.js` (passed with existing warnings)
- `bunx vitest run tests/unit/renderer/useAutoScroll.dom.test.tsx tests/unit/renderer/messageListStreaming.dom.test.tsx`
- `bunx vitest run`

### DIFF
--- a/packages/desktop/src/renderer/pages/conversation/Messages/useAutoScroll.ts
+++ b/packages/desktop/src/renderer/pages/conversation/Messages/useAutoScroll.ts
@@ -54,6 +54,7 @@ export function useAutoScroll({ messages, itemCount }: UseAutoScrollOptions): Us
   const userScrolledRef = useRef(false);
   const lastScrollTopRef = useRef(0);
   const previousListLengthRef = useRef(messages.length);
+  const previousLastMessageRef = useRef<TMessage | undefined>(messages[messages.length - 1]);
   const lastProgrammaticScrollTimeRef = useRef(0);
   const initialScrollDoneRef = useRef(false);
   const pendingAutoFollowFrameRef = useRef<number | null>(null);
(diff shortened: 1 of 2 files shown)


szafranski:
Superseded by #3532, which includes the streaming auto-scroll fix on current main.

PR #3380, opened 2026-06-20
fix(conversation): drain queued messages after switch

### LINKED ISSUE #3381: Draft: queued messages should continue after switching conversations
## Problem
When a user sends another message while a conversation is still processing, the command is queued by the active sendbox. If the user switches to another conversation before the current turn completes, the queued message can remain stuck because queue draining is tied to the mounted sendbox for the active conversation.

## Expected behavior
Queued messages should continue processing for their original conversation after the current turn completes, even if that conversation is no longer the active view.

## Observed behavior
The UI shows processing state from zero when returning, and queued/submitted work appears to make progress only while the conversation view is active.

## Proposed fix
Keep a background queue runner for conversations with pending commands after their sendbox unmounts, and wake it from turn-completed runtime events for inactive conversations.

## Related PR
- #3380

szafranski:
## Summary
- keep a detached command queue runner alive after the active sendbox unmounts
- update the runtime view from inactive conversation turn-completed events before draining queued commands
- add a regression test for queued command execution after switching away from the conversation

Fixes #3381

## Tests
- bun run format
- bun run lint (passes with existing warnings)
- bunx tsc --noEmit
- bun run i18n:types
- node scripts/check-i18n.js (passes with existing warnings)
- bunx vitest run tests/unit/conversation/runtime/conversationCommandQueueTeamUpgrade.dom.test.tsx --project dom
- bunx vitest run

### DIFF
--- a/packages/desktop/src/renderer/pages/conversation/platforms/useConversationCommandQueue.ts
+++ b/packages/desktop/src/renderer/pages/conversation/platforms/useConversationCommandQueue.ts
@@ -1,4 +1,9 @@
+import { ipcBridge } from '@/common';
 import { uuid } from '@/common/utils';
+import {
+  getConversationRuntimeViewSnapshot,
+  turnCompleted,
+} from '@/renderer/pages/conversation/runtime/conversationRuntimeViewStore';
 import { useAddEventListener } from '@/renderer/utils/emitter';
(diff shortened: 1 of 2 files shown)


szafranski:
Superseded by #3532, which includes the queue-draining fix on current main.

PR #3382, opened 2026-06-20
fix(agent): make ACP model options self-explanatory

### LINKED ISSUE #3388: [Bug/UX] ACP model dropdown doesn't show which model an option actually runs
# [Bug/UX] ACP model dropdown doesn't show which model an option actually runs

## Summary

In the ACP model selector (Claude and other ACP backends), the dropdown options are hard to scan and — more importantly — don't reveal which model/version they actually run:

- `Default (recommended)` — gives no hint that it resolves to **Sonnet 4.6**.
- `Opus` vs `Opus 4.8` — `Opus` is the alias for the latest Opus (today **Opus 4.8**), while `Opus 4.8` is the pinned `claude-opus-4-8`; the list doesn't disambiguate them.
- The qualifier (`recommended`, `1M context`) is glued onto the model name and, on the narrow header pill, can truncate the name away.

The backend already sends a per-option `description` (e.g. `Default` → "Sonnet 4.6 · Best for everyday tasks", `Opus` → "Opus 4.8 · Most capable for complex work · ~2× usage vs Sonnet"), but it never reaches the UI.

## Steps to reproduce

1. Use an ACP agent that exposes multiple models (e.g. Claude Code).
2. Open the model selector dropdown in the conversation header / sendbox.
3. Observe the option labels — there's no indication of what `Default` runs or which `Opus` is which.

## Expected

- The base model name is always the primary, prominent text; qualifiers are secondary.
- The option shows which model/version it resolves to (e.g. `Default` → Sonnet 4.6).
- Full details (usage/pricing) are available without cluttering the list.

## Actual

The raw label from the backend (`Base (qualifier)`) is rendered verbatim; the `description` field is dropped before the UI, so none of the resolved-model info is shown.

## Notes

- The naming inconsistencies themselves (`Opus` vs `Opus 4.8`, the abstract `Default`) originate **upstream** in the ACP payload — AionUi renders the labels faithfully. This issue is about AionUi's **presentation** of those labels, not about changing the backend list.
- Root cause on the AionUi side: the `description` flows through `deriveSelectOption` but is dropped one step before the UI in `useAcpModelInfo` (`available_models` map keeps only `{id, label}`).

## Proposed fix (implemented in the linked PR)

Renderer-only, no backend/DTO change:
- Keep the base name primary, demote the trailing qualifier to a secondary badge.
- Carry `description` through to the UI; show its first segment (resolved model + version) as a secondary line, and the full text on hover via a tooltip.

## Environment

- AionUi: 2.1.21
- OS: macOS (Apple Silicon)
- Backend: Claude Code (ACP)

szafranski:
Closes #3388

## Problem

In the ACP model dropdown (Claude and other ACP backends), the options were hard to read and, worse, did not say what they actually run:

- `Default (recommended)` — gives no hint that it resolves to **Sonnet 4.6**.
- `Opus` vs `Opus 4.8` — `Opus` is the alias for the latest Opus (today **Opus 4.8**), while `Opus 4.8` is the pinned `claude-opus-4-8`; the list didn't disambiguate.
- The qualifier (`recommended`, `1M context`) was glued onto the name and, on the narrow header pill, could truncate the model name away.

## Fix

Renderer-only — no backend or DTO change. The per-option `description` already flows through `deriveSelectOption`; it was being dropped one step before the UI in `useAcpModelInfo`. We now carry it through and present each option clearly:

- **Base name stays primary**; the trailing qualifier becomes a secondary badge beside it (`splitModelLabel`), so the model name is always visible.
- **Resolved model + version as a secondary line** (`modelSummaryFromDescription` — the first `·`-separated segment of the description), e.g. `Default` now shows `Sonnet 4.6`, `Opus` shows `Opus 4.8`.
- **Full description on hover** via an Arco `Tooltip` (usage/pricing detail).
- When the backend sends no description (e.g. handshake-only), the secondary line and tooltip are simply omitted.

The naming inconsistencies themselves originate upstream in the ACP payload — AionUi renders the labels faithfully; this PR is about making that payload **legible**.

## Scope

Limited to the model option list. The collapsed header pill and the thought-level group are intentionally untouched. Alias-vs-pinned labelling is deliberately left as-is (we surface the backend description verbatim rather than inventing text).

## Testing

- Unit tests for `splitModelLabel` and `modelSummaryFromDescription` (qualifier split, version suffix, parenthetical-only, empty/whitespace, null/undefined, multi-segment descriptions).
- DOM tests: option keeps the base name as its own text node + qualifier badge; resolved-model secondary line renders; full description is wired into a hover tooltip; secondary line and tooltip are omitted when no description is sent.
- `bun run format`, `bun run lint` (0 errors), `bunx tsc --noEmit`, `node scripts/check-i18n.js`, `prek` (CI replica) and full `bunx vitest run` (1577 passed) all green.

### DIFF
--- a/packages/desktop/src/common/types/platform/acpTypes.ts
+++ b/packages/desktop/src/common/types/platform/acpTypes.ts
@@ -246,7 +246,7 @@ export interface AcpModelInfo {
   /** Display label for the current model */
   current_model_label: string | null;
   /** Available models for switching */
-  available_models: Array<{ id: string; label: string }>;
+  available_models: Array<{ id: string; label: string; description?: string | null }>;
 }
 
 // ===== Permission request (session/request_permission) =====
--- a/packages/desktop/src/renderer/components/agent/AcpModelSelector.tsx
+++ b/packages/desktop/src/renderer/components/agent/AcpModelSelector.tsx
@@ -20,6 +20,7 @@ import {
   isConfigSetting,
   RuntimeSelectorCheckedItem,
   RuntimeSelectorMenuDivider,
+  RuntimeSelectorModelLabel,
   renderThoughtLevelMenuGroup,
 } from './runtimeSelectorOptions';
 
@@ -144,7 +145,7 @@ const AcpModelSelector: React.FC<{
                 }}
               >
                 <RuntimeSelectorCheckedItem selected={model.id === model_info.current_model_id}>
-                  {model.label || model.id}
+                  <RuntimeSelectorModelLabel label={model.label || model.id} description={model.description} />
                 </RuntimeSelectorCheckedItem>
               </Menu.Item>
             ))}
--- a/packages/desktop/src/renderer/components/agent/runtimeSelectorOptions.tsx
+++ b/packages/desktop/src/renderer/components/agent/runtimeSelectorOptions.tsx
@@ -5,7 +5,8 @@
  */
 
 import type { AcpConfigSetStatus, AcpDerivedOption } from '@/renderer/hooks/agent/useAcpConfigOptions';
-import { Menu } from '@arco-design/web-react';
+import { modelSummaryFromDescription, splitModelLabel } from '@/renderer/utils/model/agentLogo';
+import { Menu, Tooltip } from '@arco-design/web-react';
 import React from 'react';
 
 export const getCurrentThoughtLevelLabel = (thoughtLevel: AcpDerivedOption | null | undefined): string => {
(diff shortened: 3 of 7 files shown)


szafranski:
Render of the expected outcome - something like this. Bubble with description when hovering over given model:
obraz

szafranski:
Closing this as superseded by #3463, which merged the accepted tooltip-based ACP model description UX and closed #3388. I rechecked the remaining diff against current main; the leftover unique bits reintroduce inline/secondary model description rendering, which no longer matches the direction taken in #3463.

PR #3432, opened 2026-06-25
fix(guid): stop showing stale Codex model fallback

### LINKED ISSUE #3431: [Bug]: Codex pre-chat model picker shows stale fallback models
## Problem

The Codex model picker on the Guid/pre-chat screen can show a stale hardcoded model list before a conversation starts.

In my local profile, the Codex ACP runtime catalog contains current models such as:

- `gpt-5.5`
- `gpt-5.4`
- `gpt-5.4-mini`

After a Codex conversation starts, the runtime model state reflects that newer catalog. Before the conversation starts, the Guid picker instead shows the static Codex fallback list, including older entries such as `gpt-5.3-codex`, and does not show `gpt-5.5`.

Claude does not synthesize a stale dropdown in the same situation; it falls back to the default/CLI-model state. So the ACP pre-chat behavior is inconsistent across backends.

## Expected behavior

Codex should not show stale hardcoded models before the first message.

Either:

- the pre-chat picker should use the same ACP runtime model catalog that is available after session start, or
- if that catalog is not available through the assistant list contract yet, Codex should fall back to the same neutral default/CLI-model state as other ACP agents.

## Current behavior

The Guid assistant path reads `assistant.models`.

For auto-mode ACP assistants this list is empty, because `/api/assistants` currently populates `models` only for fixed model defaults. When the selected assistant backend is `codex`, the renderer falls back to `DEFAULT_CODEX_MODELS`.

Relevant code paths:

- `packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts`
  - `resolveInitialAssistantModel()`
  - `buildAssistantModelInfo()`
- `packages/desktop/src/common/types/codex/codexModels.ts`
- `crates/aionui-assistant/src/service.rs`
  - `definition_to_response()`

## Technical note

Recent frontend code documents that Guid/business assistant pickers should use `/api/assistants`, not `/api/agents`. That leaves a contract gap: `/api/assistants` is the source for Guid, but it does not currently expose the ACP runtime model catalog for auto-mode assistants.

The Codex-only static fallback masks that gap with stale data.

Related context:

- #2869 removed the legacy `acp.cachedModels` fallback and moved model-list ownership to backend data.
- #2718 made Guid read ACP models before first session from backend-provided catalog data.
- iOfficeAI/AionCore#425 derives ACP model catalogs from `configOptions`, including Codex-style model catalogs.

## Acceptance criteria

- Codex pre-chat picker does not show stale hardcoded models.
- Codex and Claude behave consistently when no pre-chat ACP model catalog is available.
- If the ACP runtime catalog is exposed to the assistant list path, Codex pre-chat can show current models such as `gpt-5.5` from that source of truth.

szafranski:
## Summary
- remove the Codex-only static model fallback from Guid and cron model info
- keep assistant editor model options based on the assistant catalog only
- delete the now-unused `DEFAULT_CODEX_MODELS` list

## Why
Codex could show stale frontend-defined models before a conversation started when the assistant catalog did not include models. Claude already falls back to no pre-chat model list in that case, so Codex behaved differently and could pass an outdated model override into a new session.

This PR does not add runtime ACP model catalogs to `/api/assistants`. That should be a separate backend contract change.

Fixes #3431

## Testing
- `bun run format`
- `bun run lint` (passes; existing repo warnings remain)
- `bunx tsc --noEmit`
- `bun run i18n:types`
- `node scripts/check-i18n.js` (passes; existing repo warnings remain)
- `bun run test -- tests/unit/renderer/useGuidAgentSelection.dom.test.ts tests/unit/assistants/useDetectedAgents.dom.test.ts`
- `bunx vitest run`

### DIFF
--- a/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/CreateTaskDialog.tsx
+++ b/packages/desktop/src/renderer/pages/cron/ScheduledTasksPage/CreateTaskDialog.tsx
@@ -291,7 +291,7 @@ const CreateTaskDialog: React.FC<CreateTaskDialogProps> = ({
 
   const acpCachedModelInfo = useMemo<AcpModelInfo | null>(() => {
     if (!resolvedBackend || resolvedBackend === 'gemini' || resolvedBackend === 'aionrs') return null;
-    return buildAssistantModelInfo(resolvedBackend, selectedAssistantModels);
+    return buildAssistantModelInfo(selectedAssistantModels);
   }, [resolvedBackend, selectedAssistantModels]);
 
   // Auto-pick the first available model from /api/providers when aionrs is
--- a/packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts
+++ b/packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts
@@ -34,7 +34,7 @@ export type GuidAssistantSelectionResult = {
   currentAgentModeOptions: AgentModeOption[];
 };
 
-export function resolveInitialAssistantModel(backend: string, models: string[]): string | null {
+export function resolveInitialAssistantModel(models: string[]): string | null {
   if (models.length > 0) {
     return models[0];
   }
@@ -42,7 +42,7 @@ export function resolveInitialAssistantModel(backend: string, models: string[]):
   return null;
 }
 
-export function buildAssistantModelInfo(backend: string, models: string[]): AcpModelInfo | null {
+export function buildAssistantModelInfo(models: string[]): AcpModelInfo | null {
   if (models.length > 0) {
     return {
       current_model_id: models[0],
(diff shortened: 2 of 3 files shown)


kaizhou-lab:
The direction looks correct, but #3443 has already handled the core issue: the frontend static Codex catalog has been removed, and GUID / Assistant settings no longer fall back to `DEFAULT_CODEX_MODELS`.

So this PR is now mostly cleanup: removing the unused `backend` argument from `buildAssistantModelInfo` / `resolveInitialAssistantModel`, and updating the related call sites and tests.

I checked the current diff. It is small, limited to 3 files, and I don’t see a logic issue. I’d suggest updating the PR title/description from “fix Codex fallback” to something like “remove unused backend argument after Codex fallback removal” so it reflects the current scope more accurately.

PR #3460, opened 2026-06-27
Remember selected guide assistant

### LINKED ISSUE #3457: [Feature]: Pin OpenCode as the only agent on the home "what's your plan for today?" screen and make New Chat open directly into it
### Feature Description

On the AionUi home screen (the empty-state with the heading **"Hi, what's your plan for today?"**), allow the user to set a **default selected agent/tab** — and persist the last-used one — so that opening the app pre-selects that agent (e.g. **OpenCode**) instead of always defaulting to whatever AionUi currently picks.

### Problem Statement

Every time I open AionUi, the "Hi, what's your plan for today?" screen lands on a different default agent tab, and I have to manually switch to **OpenCode** before I can start working. There is no setting I can find to make OpenCode (or my most-recently-used agent) the default, and this small manual step adds friction on every launch. This is especially frustrating for users who have standardized on a single CLI agent (OpenCode, Codex, Claude Code, etc.) as their primary workflow.

Related existing requests (#2796, #2797) are about picking the default *model/provider* for a single assistant (Ollama vs Gemini). This request is different: it is about the **default agent tab pre-selected on the home/startup screen**.

### Proposed Solution

Add a preference (under Settings, and/or driven by "remember last used") that controls which agent tab is pre-selected on the home "what's your plan for today?" screen:

- Option A (preferred): **Remember the last-used agent** on the home screen and pre-select it on the next app launch.
- Option B (explicit): A Settings dropdown, e.g. **"Default agent on home screen"**, defaulting to the last-used agent, with the ability to pin it to OpenCode (or any installed CLI agent).

Either approach removes the daily manual switch. Storing the choice alongside existing window/preferences state (e.g. next to `ui.zoomFactor` / `pet.size` in the AionUi config) would be consistent with how other UI preferences are already persisted.

### Feature Category

UI/UX Improvement

### Additional Context

- Confirmed the home screen heading string `"Hi, what's your plan for today?"` exists in the renderer, so this is specifically about that empty-state home view's agent tab selection.
- Concrete user need: I want **OpenCode** to be the default tab on that screen every time I open the app.
- Platform: macOS (AionUi 2.1.25), but this is a cross-platform UX preference.

shentry:
## Summary
- Persist the last assistant selected on the guide home screen in localStorage.
- Restore that assistant for the guide default and New Chat reset flows when it is still available.
- Add DOM hook coverage for restore, reset, and persistence behavior.

Closes #3457

## Test Plan
- ./node_modules/.bin/vitest run tests/unit/renderer/useGuidAgentSelection.dom.test.ts tests/unit/renderer/hooks/guidPage.dom.test.tsx tests/unit/renderer/hooks/guidAssistantSelectionArea.dom.test.tsx
- ./node_modules/.bin/oxfmt --check packages/desktop/src/common/config/storageKeys.ts packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts tests/unit/renderer/useGuidAgentSelection.dom.test.ts
- git diff --check origin/main..HEAD

### DIFF
--- a/packages/desktop/src/common/config/storageKeys.ts
+++ b/packages/desktop/src/common/config/storageKeys.ts
@@ -25,6 +25,9 @@ export const STORAGE_KEYS = {
 
   /** Language preference / 语言偏好 */
   LANGUAGE: 'aionui_language',
+
+  /** Last assistant selected on the guide home screen / 引导首页上次选择的助手 */
+  GUID_LAST_SELECTED_ASSISTANT_ID: 'aionui_guid_last_selected_assistant_id',
 } as const;
 
 export type StorageKey = (typeof STORAGE_KEYS)[keyof typeof STORAGE_KEYS];
--- a/packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts
+++ b/packages/desktop/src/renderer/pages/guid/hooks/useGuidAssistantSelection.ts
@@ -5,6 +5,7 @@
  */
 
 import { assistantRuntimeKey, isAionrsAssistant, type Assistant } from '@/common/types/agent/assistantTypes';
+import { STORAGE_KEYS } from '@/common/config/storageKeys';
 import type { AcpModelInfo } from '../types';
 import type { AgentModeOption } from '@/renderer/utils/model/agentTypes';
 import {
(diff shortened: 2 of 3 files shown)


kaizhou-lab:
Checked this against #3457 and the current codebase persistence pattern.

This PR covers the original \"remember last-used assistant\" behavior, but it persists the selection in `localStorage`. For this repo, UI preferences like `ui.zoomFactor` / `pet.size` go through `configService` and `/api/settings/client`, so this should use backend client settings as the authoritative store instead of browser-local storage.

I opened #3468 with the same remember-last-selected behavior backed by `guid.lastAssistantId` in client settings. It also restores the saved assistant in the New Chat / Guid reset path, which otherwise requires a refresh before the remembered selection takes effect.

One scope note: #3457 was later clarified to include pinned-only-agent behavior, hiding other tabs, and New Chat directly opening that agent. The remember-last-used change is only the first part of that request, so I would avoid closing #3457 until that broader pinned-agent scope is either implemented or split out.

PR #3463, opened 2026-06-28
fix(agent): show ACP model descriptions

### LINKED ISSUE #3388: [Bug/UX] ACP model dropdown doesn't show which model an option actually runs

shentry:
## Summary
- Preserve ACP model option descriptions from runtime payloads and config options
- Render descriptions in desktop, Guid, and mobile model selectors
- Add regression coverage for model descriptions flowing through the selector data path

Fixes #3388

## Verification
- PATH="/Users/zxy/Library/Caches/aionui/runtime/bun-1.3.13-fc0b4cae13a9:$PATH" just push -u fork codex/issue-3388-acp-model-descriptions
- CDP manual check: Claude Code model dropdown contains descriptions such as "Use the default model..." and "Custom Opus model..."


kaizhou-lab:
Updated this PR to match the latest UX direction.

The implementation is now different from the original inline-description approach: model descriptions are no longer rendered as a second line in the option list, because mixed options with and without descriptions made the dropdown rows visually inconsistent. Descriptions are now shown on hover via tooltips.

Coverage was also expanded beyond models:
- ACP model selector: model descriptions show in option tooltips.
- Guid/cron model selector: ACP model descriptions show in option tooltips.
- Assistant settings default model selector: model descriptions show in option tooltips.
- Agent mode selector: mode descriptions show in option tooltips.
- Runtime thought-level options: descriptions show in option tooltips.
- Mobile action sheet keeps inline descriptions, since there is no hover interaction on mobile.

Pushed commit: 3cbcba3ac24aa762d5b887f29de98b3e58f8d0d1

Verification passed via `just push -u shentry codex/issue-3388-acp-model-descriptions`:
- lint / format / typecheck
- i18n checks (existing warnings only)
- full Vitest: 256 passed, 1 skipped files; 1887 passed, 4 skipped tests

PR #3482, opened 2026-07-01
feat(conversation): surface session skills in slash command menu

IceyLiu:
## What

Skills mounted on a conversation were only reachable through the `+` menu. Now they also show up in the `/` autocomplete list: each loaded skill is offered as an insert-style template (`/name `), so typing `/` surfaces the session's skills and selecting one fills the input for the user to add arguments before sending.

## Why

Users expected to invoke conversation-loaded skills via slash, matching the CLI's slash-trigger behavior. Previously `/` only listed built-in and ACP-agent commands.

## How

- `SlashCommandSource` gains a `skill` value (labeling only — no logic branches on it).
- New pure helpers `buildSkillSlashCommands` / `mergeSlashCommands` in `common/chat/slash/`, with merge priority **builtin > ACP > skills** (existing names always win, no duplicates).
- `SendBox` builds skill commands from the conversation's `loadedSkills` snapshot and reuses the global skills index (shared SWR key `skills-index`) for descriptions.
- i18n key `conversation.skills.slashHint` added across all 10 locales as the description fallback.

## Scope / risk

Frontend + renderer only. **No aioncore / IPC / backend changes.** Reuses existing data (loadedSkills snapshot + skills index) and the existing slash-menu mechanism.

## Testing

- White-box: new unit tests for `buildSkillSlashCommands` / `mergeSlashCommands` (priority, dedup, fallback). Full suite green (1845 passed).
- tsc + lint clean, i18n types in sync.
- Manual: verified in dev that typing `/` in a conversation with loaded skills surfaces those skills in the menu.


IceyLiu:
close #3455

PR #3602, opened 2026-07-16
feat(guid): add Ollama Launch model selector for compatible ACP agents

ChristianLuciani:
⏸️ Paused — see discussion in #3797


ChristianLuciani:
Closing this PR. Ollama already works as a `custom` platform with a configurable `base_url` under Settings -> Platforms, so this UI shortcut is redundant with existing configuration. Keeping issue #3797 open in case a UI-level selector is wanted. Thanks for the consideration.

PR #3659, opened 2026-07-21
fix(chat): bound HorizontalFileList to conversation width to prevent overflow

### LINKED ISSUE #3656: Message with many attached files stretches the chat area and creates a horizontal scrollbar (no wrapping)
## Summary

The chat conversation area is supposed to be width-adaptive, with message content wrapping to fit. But when a user message contains **many attached files**, the attachment row **does not stay within the conversation width** — it stretches the whole middle conversation area open and produces a page-level horizontal scrollbar.

## Steps to reproduce

1. In a conversation, attach/paste a large number of files to a single message and send it.
2. Observe the sent user message's attachment row.

**Expected:** the attachment strip stays within the conversation/message width; overflowing files scroll **inside** the strip (it already has left/right scroll buttons for this).
**Actual:** the attachment strip expands to its full content width, pushing the conversation area wider than the viewport and creating a horizontal scrollbar for the whole chat area.

## Investigation notes (for contributors)

The multi-file attachment strip is rendered by `HorizontalFileList`:

- `packages/desktop/src/renderer/components/media/HorizontalFileList.tsx`
  - The scroll container is `flex items-center gap-8px overflow-x-auto ...` (line ~118-120) and is *designed* to scroll horizontally internally (it computes `scrollWidth > clientWidth` and shows left/right buttons).
  - **Problem:** the component has no effective max-width / `min-w-0` constraint, so instead of scrolling within a bounded width, its own width grows with the content and stretches the ancestor layout.
- `packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx`
  - Multi-file case renders `<HorizontalFileList>` (line ~191). The wrapping `<div className='mt-6px ...'>` (line ~184) and the message container may need `min-w-0` / width bounding so the flex child can shrink.

**Suggested fix:** bound the horizontal list to the available conversation width (e.g. `w-full max-w-full min-w-0` on the scroll container, and ensure the ancestor chain has `min-w-0` so a flex item can shrink below its content size). The internal horizontal scroll + scroll buttons should then do their job instead of the whole chat area widening. The single-file path (`FilePreview`) is fine — this is specifically the multi-file `HorizontalFileList` case.

## Scope

Pure front-end CSS/layout (`packages/desktop/src/renderer/`). No IPC / aioncore changes. Likely 1 file (`HorizontalFileList.tsx`), possibly a small `min-w-0` tweak in `MessageText.tsx`.

taekchef:
## Summary

A user message with **many attached files** stretched the whole conversation area wider than the viewport and produced a page-level horizontal scrollbar. The multi-file \`HorizontalFileList\` is *designed* to scroll horizontally internally (it already computes \`scrollWidth > clientWidth\` and shows left/right scroll buttons), but it had no effective width constraint, so instead of scrolling inside a bounded width, its own width grew with the content and pushed the ancestor layout open.

The single-file \`FilePreview\` path is unaffected — this only impacts the multi-file \`HorizontalFileList\` case.

## Fix

Bound the list to the available conversation width so the flex child can shrink and the existing inner horizontal scroll + scroll buttons do their job:

- \`HorizontalFileList.tsx\`
  - Root \`relative\` → \`relative min-w-0 max-w-full\`
  - Scroll container → added \`w-full max-w-full min-w-0\`
- \`MessageText.tsx\` — wrapping \`<div>\` around the multi-file list → added \`min-w-0 max-w-full\` so the ancestor chain allows the flex item to shrink below its content size

## Test

Added \`tests/unit/renderer/horizontalFileList.dom.test.tsx\` — renders \`<HorizontalFileList>\` and asserts the root and scroll container carry the width-bounding classes (\`min-w-0\` / \`max-w-full\` / \`overflow-x-auto\`) to guard against regression.

Closes #3656.

## Checklist

- [x] Front-end CSS/layout only — no IPC / aioncore changes
- [x] \`bunx tsc --noEmit\` — no type errors
- [x] \`bun run lint:fix\` — 0 errors
- [x] \`bun run format\` — applied
- [x] New unit test passes

### DIFF
--- a/packages/desktop/src/renderer/components/media/HorizontalFileList.tsx
+++ b/packages/desktop/src/renderer/components/media/HorizontalFileList.tsx
@@ -113,11 +113,11 @@ const HorizontalFileList: React.FC<HorizontalFileListProps> = ({ children }) => 
   };
 
   return (
-    <div className='relative'>
+    <div className='relative min-w-0 max-w-full'>
       {/* 横向滚动容器，隐藏滚动条 */}
       <div
         ref={scrollContainerRef}
-        className='flex items-center gap-8px overflow-x-auto overflow-y-hidden scrollbar-hide pt-5px pb-5px'
+        className='flex items-center gap-8px overflow-x-auto overflow-y-hidden scrollbar-hide pt-5px pb-5px w-full max-w-full min-w-0'
         style={{
           scrollbarWidth: 'none', // Firefox
           msOverflowStyle: 'none', // IE/Edge
--- a/packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx
+++ b/packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx
@@ -241,7 +241,7 @@ const MessageText: React.FC<{ message: IMessageText; showCopyRow?: boolean }> = 
           </div>
         )}
         {files.length > 0 && (
-          <div className={classNames('mt-6px', { 'self-end': isUserMessage })}>
+          <div className={classNames('mt-6px min-w-0 max-w-full', { 'self-end': isUserMessage })}>
             {resolvedFiles.length === 1 ? (
               <div className='flex items-center'>
                 <FilePreview path={resolvedFiles[0]} onRemove={() => undefined} readonly />
--- a/tests/unit/renderer/horizontalFileList.dom.test.tsx
+++ b/tests/unit/renderer/horizontalFileList.dom.test.tsx
@@ -1,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2025 AionUi (aionui.com)
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Regression test for #3656: a message with many attached files stretched the
+ * chat area and produced a page-level horizontal scrollbar. The fix bounds the
+ * HorizontalFileList to the available conversation width so the inner strip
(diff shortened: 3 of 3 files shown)


IceyLiu:
@taekchef Code review passed ✅. Could you add **before/after screenshots** showing the conversation area with many attached files — specifically that it no longer stretches horizontally and the scroll stays contained within the message?

IceyLiu (approved):
验收通过。多附件横向溢出问题已修复，flex 容器加 min-w-0 / max-w-full 后滚动正常，气泡不再被撑破。改动最小且附带单元测试。

taekchef:
Added the requested before/after evidence.

Both captures use the same isolated setup: branch build, 1024 x 768 viewport, one user message, and the same 12 attached files.

Measured result:
- **Before:** the message column had 461 px available, but the attachment row expanded to 2824 px (`clientWidth === scrollWidth`). The first attachment was positioned at x = -2074, so only the tail of the list was visible.
- **After:** the wrapper and attachment viewport stay at 461 px while the 2824 px content remains inside the row as horizontal scroll. The first attachment starts inside the message at x = 289.
- The row's right control was exercised successfully: inner `scrollLeft` moved from 0 to 200 without changing the message or page width.

Screenshots:
1. Before, the row is stretched and earlier files are pushed off-screen
2. After, the row is contained in the message and exposes its internal scroll control
pr3659-before
pr3659-after

PR #3666, opened 2026-07-21
feat(team): show running state in sidebar

### LINKED ISSUE #3646: feat(team): show loading spinner on team sidebar item while team is running
## Background

When an image generation conversation is running, its entry in the left sidebar shows a spinning loader instead of the leading icon. Teams have no equivalent visual indicator — users cannot tell at a glance whether a team is still running.

## Desired behaviour

While a team has at least one non-terminal run in progress (`TeamRunStatus` is `'accepted'`, `'running'`, or `'cancelling'`), its sidebar entry (both expanded and collapsed states) should display a `<Spin>` in place of the leading icon, matching the existing pattern in `ConversationRow`.

The spinner disappears as soon as the team reaches a terminal status (`'completed'`, `'cancelled'`, or `'failed'`).

## Implementation hint

`TeamRunStatus` and the relevant IPC run-event stream already exist. A new `useSiderTeamRunning(teams)` hook — modelled after `useSiderTeamBadges` — should subscribe to team run events and expose an `isTeamRunning(teamId): boolean` helper. `TeamSiderSection` then uses that helper to conditionally render `<Spin size={16} />` instead of the `<Peoples>` icon, exactly as `ConversationRow` does with `isConversationGenerating`.

## Acceptance criteria

- [ ] A running team's sidebar entry shows a spinner in place of the group icon
- [ ] The spinner disappears when the run reaches a terminal state
- [ ] Both the expanded list view and the collapsed icon view show the spinner
- [ ] No visual change for teams that are idle

eyaeya:
## Description

TeamSiderSection currently knows about the team list and pending-permission badge counts, but it has no source of run lifecycle state. A team therefore keeps the same Peoples icon while work is accepted, running, or cancelling, including when the user navigates away from the team page.

This change adds a sidebar-scoped run-state hook and uses it in both expanded and collapsed layouts.

The hook:

- reconciles each visible team with the authoritative getRunState endpoint on load;
- subscribes to accepted, started, updated, completed, cancelled, and failed lifecycle events;
- reconciles again after a WebSocket reconnect so events missed while offline do not leave the icon stale;
- tracks active run IDs instead of one boolean, so a late terminal event for an older run cannot hide a newer active run;
- rejects out-of-order snapshots and serially retries when a live event arrives during reconciliation;
- uses a global request sequence to prevent an old response from being accepted if a team is removed and later re-added with the same ID;
- removes state for deleted teams and cleans up all seven listeners on unmount.

While a team is active, its sidebar icon becomes an Arco Spin in both layouts. The existing permission badge stays visible. In the expanded layout, the hover pin overlay is suppressed while the spinner is present; idle teams retain their current icon and pin behavior.

One existing test file is moved unchanged into the new renderer/layout test directory. The move is path-only and keeps the already oversized renderer test directory from gaining another direct child.

## Reproduction

On current main:

1. Create or open a team with at least one runnable assistant.
2. Send a message that keeps the team active long enough to navigate away.
3. Open another page while the run is accepted or running.
4. Observe the team entry in the expanded sidebar.
5. Collapse the sidebar and inspect the same entry again.

Before this change, both entries keep the static team icon, so there is no indication that the background team is still working. After this change, both entries show a 16px spinner until the authoritative run reaches a terminal state.

The regression suite also reproduces the less visible reconnect case: retain a stale run A, begin reconciliation, receive and finish run B while that request is in flight, then resolve the stale snapshot. The follow-up reconciliation must clear run A instead of leaving the spinner stuck.

## Related Issues

- Fixes #3646

## Type of Change

- [ ] fix — Bug fix
- [x] feat — New feature (non-breaking change which adds functionality)
- [ ] perf — Performance improvement
- [ ] refactor — Code restructuring (no behavior change)
- [ ] Breaking change
- [ ] docs — Documentation update

- [x] This PR contains exactly one feature: reflecting active team execution in its sidebar entry
- [x] The PR title follows Conventional Commit format: type(scope): subject (English)

## Local Checks (Rule 3)

- [x] bun run format:check — all 1,740 files conform
- [x] bun run lint -- --quiet — 0 errors
- [x] bunx tsc --noEmit — no type errors
- [x] bunx vitest run — 309 files passed, 1 skipped; 2,386 tests passed, 5 skipped
- [x] i18n validated (bun run i18n:types and node scripts/check-i18n.js)
- [x] No user-facing text was added

## Regression Coverage

The focused suite contains 36 passing tests across three files and covers:

- every active and terminal lifecycle event;
- multiple run IDs and late terminal events;
- isolation between teams;
- initial hydration and reconnect reconciliation;
- stale snapshots, response reordering, and live-event races;
- team deletion and same-ID re-addition;
- failed snapshots and unmount cleanup;
- spinner versus idle icon behavior in expanded and collapsed sidebars;
- running, idle-pinned, and idle-unpinned pin presentation.

## Runtime Verification

- [ ] Verified on macOS
- [ ] Verified on Windows
- [ ] Verified on Linux

## Screenshots

Not included. The two layout branches are covered with DOM tests that assert the spinner size and the absence of the idle icon while a run is active.

## Additional Context

Initial load and reconnect perform one lightweight run-state request per visible team. Requests are independent, and stale responses cannot overwrite newer lifecycle events. No backend event, API contract, team-run state machine, badge count, or locale resource is changed.


IceyLiu:
@eyaeya Code review passed ✅. Could you add **screenshots** showing the loading spinner on the team sidebar item while a team run is in progress — both the expanded list view and the collapsed icon view?

eyaeya:
While producing stable-state Electron screenshots, I found a product-level conflict in the collapsed case. The expanded spinner renders correctly, but the desktop sider intentionally settles at `collapsedWidth=0` following #2859, so the entire collapsed icon view — including the new team spinner branch — is clipped after the transition finishes.

Changing desktop collapse to a 64px icon rail makes the spinner visible, but it also reverses the #2859 behavior globally and exposes existing full-width content (for example, the empty conversation-history state) in the narrow rail. I do not want to hide that regression behind a transition-frame screenshot.

Could you confirm which direction you prefer?

1. Restore and fully polish the 64px desktop icon rail in this PR (while keeping mobile at 0px), including the surrounding collapsed-content fixes and regression coverage.
2. Keep the current full-hide behavior and revise the collapsed-icon acceptance criterion.

The expanded-state Electron capture is ready; I will attach the final matched pair once the intended collapsed behavior is clear.

IceyLiu (approved):
Hands-on verified ✅

App starts correctly with no regressions. The implementation looks solid — when a team run is active, the sidebar icon swaps to a Spin loader; when the run completes it reverts to the Peoples icon. Both collapsed and expanded sidebar views are covered, and the hook correctly handles live events, reconnect reconciliation, and stale-request guards. Ready to merge.

IceyLiu:
@eyaeya Reviewed and approved ✅ Great work on the running state hook — the event handling and reconciliation logic is thorough. Ready to merge!

PR #3677, opened 2026-07-21
fix(workspace): allow Changes tab list to scroll on large changesets

### LINKED ISSUE #3652: Workspace Changes tab does not scroll when there are many changed files
## Summary

In the Workspace side panel, the **Changes** tab cannot scroll when there are many changed files. With a large changeset (e.g. 133 files), the list overflows the visible area but the user cannot scroll down — the bottom file entries and the "Staged changes" group are cut off and unreachable.

## Steps to reproduce

1. Open a conversation whose workspace is a git repo with many changed files (100+).
2. Open the Workspace panel and switch to the **Changes** tab.
3. Try to scroll the file list down to the bottom.

**Expected:** the list scrolls and all changed files + the staged/unstaged groups are reachable.
**Actual:** the list is stuck; the lower entries are cut off and cannot be scrolled into view.

## Investigation notes (for contributors)

This looks like a **nested scroll-container / height conflict** in the layout, not missing overflow. There appear to be two stacked scroll containers:

- `packages/desktop/src/renderer/pages/conversation/Workspace/index.tsx` wraps the Changes tab in `<FlexFullContainer containerClassName='overflow-y-auto'>` (its inner div is `absolute size-full overflow-y-auto`).
- `packages/desktop/src/renderer/pages/conversation/Workspace/components/FileChangeList.tsx` root is `flex flex-col size-full`, and its inner list area is again `flex-1 overflow-y-auto` (line ~406).
- `packages/desktop/src/renderer/components/layout/FlexFullContainer.tsx` provides the `flex-1 relative min-h-0` + `absolute size-full` wrapper.

Please check whether the two nested `overflow-y-auto` layers (and `size-full` / `flex-1` height resolution) are fighting each other, and settle on a single well-bounded scroll container so the full list scrolls. Note the `files` tab (Tree) in the same file scrolls correctly with just the outer container — that's a useful comparison point.

## Scope

Pure front-end CSS/layout (`packages/desktop/src/renderer/`). No IPC / aioncore / data-layer changes expected. Likely a 1-2 file fix.

gaursomya777-dev:
## Summary
Fixes nested scroll-container conflict that prevented the Workspace **Changes** tab from scrolling when many files changed.

## Changes
- Remove competing `overflow-y-auto` from the Changes tab outer `FlexFullContainer` (keep `min-h-0` height chain)
- Add `min-h-0` on `FileChangeList` root + list region so flex children can shrink
- List region is the sole scroll owner; toolbar stays pinned

## Test plan
- [ ] Open a workspace with 100+ changed files
- [ ] Switch to Changes tab
- [ ] Scroll through unstaged/staged groups — all entries reachable
- [ ] Files tree tab still scrolls as before

Fixes #3652


gaursomya777-dev:
Closing in favor of the assignee IceyLiu selected on the linked `bonus` issue.

Sorry for the noise — I opened this before the assignment landed. Happy to help review or pick up a different unassigned bonus issue instead.

PR #3727, opened 2026-07-24
fix(conversation): long URL / unbroken string in a user message overflows the bubble

### LINKED ISSUE #3724: fix(conversation): long URL / unbroken string in a user message overflows the bubble
## Bug description

When a user sends a message containing a long URL or file path (a long run of characters **without spaces**), the text does not wrap. The message bubble is stretched past the window edge and the beginning of the string gets clipped off-screen.

This is the same family as the earlier "too many files in a row overflow" fix, but for long unbroken text rather than file chips.

## Steps to reproduce

1. Open any conversation.
2. Send a message containing a long path/URL with no spaces, e.g.
   `/var/folders/gd/6bb7q8jd1ll0g17q5gly4flw0000gn/T/aionui/0265f4a8/image-xxxxxxxxxxxxxxxxxx`
3. Look at the sent user-message bubble.

**Expected:** the long string wraps inside the bubble; the bubble stays within the window width.
**Actual:** the string stays on one line, stretches the bubble past the window, and the left part is clipped.

## Root cause (for contributors)

`packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx` (~line 277) renders the user-message text with:

```jsx
<div className='whitespace-pre-wrap break-words' data-testid='message-text-content'>
```

`break-words` (`overflow-wrap: break-word`) only breaks at whitespace/word boundaries. A long no-space string is treated as a single "word" with no break opportunity, so it never wraps.

## Suggested fix

Switch to `overflow-wrap: anywhere` (preferred — normal words still break at word boundaries, but a long unbroken run breaks by character when needed), or `word-break: break-all`. Prefer `anywhere` so normal CJK/English paragraphs aren't affected.

## Acceptance criteria

1. A long no-space URL/path in a **user message** wraps inside the bubble and never pushes the bubble past the window width.
2. Normal mixed CJK/English paragraphs render unchanged (no per-character breaking of ordinary text) — this is why `anywhere` is preferred over `break-all`.
3. Check the AI-reply side (Markdown-rendered) for the same overflow and fix it if present (the screenshot is a user message, but worth verifying).
4. Add/adjust a focused test.

## Scope

Renderer-only, `MessageText.tsx` (likely a one-line className change). No main-process / AionCore changes.

GoldenDecade:
## What & Why

Fixes #3724.

When a user sends a message containing a long unbroken string (a URL or file path with no spaces), the text did not wrap: the message bubble stretched past the window edge and the start of the string was clipped off-screen.

**Root cause:** the user-message plain-text branch in `MessageText.tsx` used `whitespace-pre-wrap break-words`. `overflow-wrap: break-word` only breaks at word boundaries and does **not** shrink min-content, so a no-space run is treated as one unbreakable word and stretches the bubble.

**Fix:** switch that one class to `overflow-wrap: anywhere`, which breaks a long unbroken run by character **and** shrinks min-content so the bubble stays within the row. Normal CJK/English paragraphs are unaffected (they still break at word boundaries). This matches the pattern already used on the AI-reply side (`ShadowView.tsx` `.markdown-shadow-body`), which is why the Markdown path already wrapped correctly and needed no change.

## Scope

Renderer-only, one-line className change in `MessageText.tsx`. No main-process / AionCore changes.

## Testing

- Added a focused regression test asserting a long-path user-message bubble carries `overflow-wrap: anywhere` and no longer uses `break-words`.
- `messageText.dom.test.tsx`: 20/20 pass.
- Full suite: 2577 passed, 0 failed.
- `tsc --noEmit`, lint, format, i18n checks all pass.

## Checklist

- [x] Follows CONTRIBUTING.md
- [x] Renderer/main process boundary respected
- [x] New user-facing text uses i18n (n/a — no new text)
- [x] Tests added/updated and passing
- [x] No AI signature in commits

### DIFF
--- a/packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx
+++ b/packages/desktop/src/renderer/pages/conversation/Messages/components/MessageText.tsx
@@ -274,7 +274,7 @@ const MessageText: React.FC<{ message: IMessageText; showCopyRow?: boolean }> = 
         >
           {/* JSON 内容使用折叠组件 Use CollapsibleContent for JSON content */}
           {shouldRenderPlainText ? (
-            <div className='whitespace-pre-wrap break-words' data-testid='message-text-content'>
+            <div className='whitespace-pre-wrap [overflow-wrap:anywhere]' data-testid='message-text-content'>
               {text}
             </div>
           ) : json ? (
--- a/tests/unit/chat/messageText.dom.test.tsx
+++ b/tests/unit/chat/messageText.dom.test.tsx
@@ -255,6 +255,19 @@ describe('MessageText attachment paths', () => {
     expect(content.parentElement?.className).not.toContain('max-w-780px');
   });
 
+  it('wraps a long unbroken url/path in a user message so the bubble never overflows', () => {
+    const longPath = '/var/folders/gd/6bb7q8jd1ll0g17q5gly4flw0000gn/T/aionui/0265f4a8/image-xxxxxxxxxxxxxxxxxx';
+
+    renderMessageText(longPath, { position: 'right' });
+
+    const content = screen.getByTestId('message-text-content');
+    // overflow-wrap: anywhere breaks a no-space run by character AND shrinks
+    // min-content, so the bubble stays within the row. break-words does neither.
+    expect(content.className).toContain('[overflow-wrap:anywhere]');
+    expect(content.className).not.toContain('break-words');
+    expect(content).toHaveTextContent(longPath);
+  });
+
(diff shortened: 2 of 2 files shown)


IceyLiu (approved):
Verified locally (commit 46d91ac75). The one-line fix — swapping `break-words` for `overflow-wrap: anywhere` on the user-message plain-text branch — correctly wraps long unbroken URLs/paths inside the bubble instead of stretching it past the window edge, while normal CJK/English paragraphs still break at word boundaries. Added test covers the wrap behavior; 20 messageText dom tests pass and tsc is clean. Approving. Thanks @GoldenDecade!

PR #3743, opened 2026-07-26
feat(sendbox): copy sent prompt to clipboard when enabled

### LINKED ISSUE #3540: [Feature]: place prompt on clipboard before sending
### Feature Description

Imagine this scenario:
1. You type a great prompt in the input field that took you some time to compose.
2. You hit send
3. You see the words "processing" 
4. After a while you come to the realization that something is not right.
5. You refresh the screen and now you're being prompted to login.
6. That well crafted prompt that took you time to compose...all gone with no chance to recover it.
7. Now you have to go into flow mode again to retype the prompt.
8. This time around, you're wiser, so you select it all and hit cmd+C to copy it to the clipboard, just in case, and then you hit send.
9. If the system crashes again, no worries, you simply paste back that last prompt and you're good.
10. You do this several times and  it only makes sense that AionUI supports this capability directly.

If you could have a setting that allows us to choose to have the app save the most recent prompt in the clipboard, that'd be great.

### Problem Statement

1. You type a great prompt in the input field that took you some time to compose.
2. You hit send
3. You see the words "processing" 
4. After a while you come to the realization that something is not right.
5. You refresh the screen and now you're being prompted to login.
6. That well crafted prompt that took you time to compose...all gone with no chance to recover it.

### Proposed Solution

If you could have a setting that allows us to choose to have the app save the most recent prompt in the clipboard, that'd be great.

### Feature Category

UI/UX Improvement

### Additional Context

_No response_

shentry:
# Pull Request

## Description

Adds an opt-in **Copy Prompt to Clipboard on Send** preference under **Settings → System** (`input.copyPromptOnSend`, default off). When enabled, every message the user sends is also placed on the system clipboard, so a crash or a forced re-login cannot destroy a long, carefully composed prompt.

**Implementation notes**

- New config key `input.copyPromptOnSend` in `ConfigKeyMap` only. It is deliberately **not** added to `IConfigStorageRefer` / `ALL_LEGACY_KEYS` — `configService` is backend-backed (`PUT /api/settings/client`), and those two lists exist to migrate values out of the legacy local config file. A brand-new key has no legacy value to migrate.
- Logic lives in a small dedicated module, `renderer/utils/chat/promptClipboard.ts`, so the decision is unit-testable independently of the 1600-line `SendBox`.
- Reuses the existing `renderer/utils/ui/clipboard.ts` `copyText()` helper, which already handles the WebUI-over-HTTP case where `navigator.clipboard` is unavailable and it must fall back to `execCommand`.

Three behavioral decisions worth calling out for review:

1. **Backs up `input`, not `finalMessage`.** `finalMessage` may contain the reply-quote blockquote and a full DOM-snippet HTML dump. Putting kilobytes of generated HTML on the clipboard would defeat the purpose — the user needs the text they typed.
2. **Blank drafts never touch the clipboard.** An attachment-only send has no composed text worth preserving, and overwriting the clipboard with `''` would destroy the user's own copied content — the exact opposite of the safety net this feature is for.
3. **Clipboard failures are swallowed.** A permission error must never block or delay the send.

Both send paths are covered: the normal send and the BTW (side-question) path.

## Related Issues

- Closes #3540

## Type of Change

## Local Checks (Rule 3)

- [x] `bun run lint` — no lint errors (exit 0; the 846 reported items are pre-existing repo-wide warnings, not errors)
- [x] `bunx vitest run` — tests pass (322 files / 2581 tests green, including 5 new cases)
- [x] i18n validated (`bun run i18n:types` + `node scripts/check-i18n.js`) — `i18n-keys.d.ts` regenerated and committed

All of the above also ran again as part of `just push` (`lint-strict` → `fmt-check` → `typecheck` → `i18n-check` → `test`), which passed before the push.

## Runtime Verification

- [ ] Verified on macOS
- [ ] Verified on Windows
- [ ] Verified on Linux

**Honest disclosure:** I have **not** launched the packaged app to verify this at runtime on any platform — the change is covered by unit tests and the full static/test gate, but the rendered Settings row and the actual clipboard write have not been observed by eye. Happy to add runtime verification or a screenshot if you'd like that before merge.

## Screenshots

None — see the runtime-verification note above.

## Additional Context

**i18n:** `settings.copyPromptOnSend` and `settings.copyPromptOnSendDesc` were added to all 13 locales in `supportedLanguages`. Translations were written per-language rather than machine-converted; in particular zh-TW uses 剪貼簿 / 當機 / 登出 rather than a direct zh-CN transliteration.

**Placement:** I put the toggle in `SystemModalContent` alongside `saveUploadToWorkspace` / `autoPreviewOfficeFiles`, following the same optimistic-update-with-rollback pattern. If you'd rather this live somewhere else (e.g. grouped with other input preferences), it's a small move — happy to relocate it.


shentry:
Pushed a follow-up commit adding four focused tests for the new Settings row (default-off, restoring a persisted value, persisting a toggle, and rolling the optimistic flip back when the backend write fails), which covers the `SystemModalContent` lines Codecov flagged.

Two lines remain uncovered by design: the `backupPromptToClipboard(...)` call sites in `SendBox/index.tsx`. Both are single delegating statements into `promptClipboard.ts`, which is at 100% line coverage. The existing SendBox suites stub the component rather than rendering it, so covering those two lines would mean standing up a new full-render harness for a 1600-line component — out of scope for an atomic PR. Happy to do it in a follow-up if you would rather have it.

IceyLiu:
Thanks a lot for the clean, well-documented implementation — the dedicated testable module and the careful i18n work are genuinely nice.

After review, though, we've decided not to take this one: AionUi already lets users recover previously sent prompts via the **Up/Down arrow keys** in the input box, so a prompt isn't actually lost on refresh/re-login, and a clipboard write carries the downside of overwriting whatever the user had copied. We're closing the underlying issue (#3540) as not-needed, so we'll close this PR alongside it.

This isn't a reflection on the code quality — please do pick up one of the open `bonus` issues, we'd love to have your next PR count toward the campaign. Thanks again!

PR #3826, opened 2026-08-03
feat(preview): add agent-controllable in-app browser over a single-target CDP bridge

IceyLiu:
## Summary

Adds an in-app browser that both the user and the agent can drive. Web pages open as tabs in the existing preview panel, sharing one persistent session so a sign-in done by hand carries over to whatever the agent does next.

The agent reaches it through a new built-in MCP server (`aionui-browser`) and a **single-target CDP bridge** that exposes only the browser webview. Nothing is written to config or the database: the bridge's port and access token travel down the process-inheritance chain (Electron main → aioncore → MCP), so multiple app instances never cross wires.

No changes to the Rust core — this is renderer, IPC, and Electron-main only.

## Why the bridge exists

The obvious way to let an agent drive Chromium is `--remote-debugging-port`. That switch is **application-wide and has no per-target ACL**: enabling it exposes *every* WebContents — including the main window with its preload bridge — to any local process, with no authentication. Since agent browser control defaults to on, that was the default posture, and it would have shipped that way.

The replacement narrows the surface to exactly one target:

- **One target only.** `cdpTargetProtocol.ts` presents a browser that happens to have exactly one tab. `chrome-devtools-mcp` connects via puppeteer's `browserURL`, which treats the endpoint as a whole browser and expects the `Target` domain, but Electron's `webContents.debugger` is page-level and cannot answer those. The bridge answers `Target.*` locally and forwards everything else. This works because Electron's debugger supports `sessionId` in both directions, which is what puppeteer's flatten mode needs.
- **Never the main window.** Attach refuses any WebContents whose `getType()` is not `webview`.
- **Authenticated.** OS-assigned port on `127.0.0.1`, with a random token required on the WebSocket upgrade, compared in constant time.
- **No silent fallback.** If the MCP cannot read both port and token it exits. Otherwise `chrome-devtools-mcp` spawns its own Chrome, and the agent drives an invisible browser while the side panel sits idle — a failure mode that is very hard to diagnose.

Refusals are explicit rather than silent no-ops. `Target.createTarget` errors instead of returning the existing target: pretending success would leave the agent driving the old page while believing it had a new one. `Browser.close` is refused because it would terminate the app.

## Trade-offs

- The agent cannot open new tabs over CDP, and additional browser contexts are refused.
- With several tabs open the bridge attaches to the active webview only, re-pointing on switch — exposing one page rather than all of them.
- `debugger.attach()` conflicts with DevTools open on the same view; the failure is surfaced with a readable reason.

## What's included

**Browser feature**

- `browser` preview content type, plus a shared session partition and MCP name in `common/` so the main and renderer processes cannot drift apart
- Built-in browser MCP server pointing `chrome-devtools-mcp` at the bridge, registered idempotently on startup and pinned to a fixed version rather than `@latest`
- Browser tabs in `PreviewContext` with a bounded tab limit, so the agent cannot open tabs without end
- Always-mounted browser layer: tabs stay alive and only toggle visibility, because unmounting would reload the page and interrupt whatever the agent is mid-way through
- Activity indicator on tabs the agent is currently driving, plus a one-time notice the first time it happens
- Browser data management in system settings, now clearing the HTTP and auth caches as well — `clearStorageData` alone does not, while the settings copy promises it does

**Settings**

- A production-visible switch for agent browser control, defaulting to on. The only existing switch lived in developer settings, a section that returns `null` when packaged, so production users had no way to turn off a capability that was already enabled. Copy states what the agent can do, the bounds of it, and what turning it off costs.

**Fixes**

- Back/forward buttons did nothing. `canGoBack` was tracked in hand-rolled URL stacks that redirects and prop echoes could desync from the webview's real history, producing an enabled-but-dead button. Both now delegate to the webview's own history.
- Redirect chains could ping-pong the address until the renderer process died. Only the most recent internal navigation was remembered, so an earlier target's late-arriving echo looked like an external change and re-triggered the redirect. A bounded set of pending echoes is now tracked, each consumed once.
- The browser toolbar overflowed the preview panel: `.toolbar-input` had `width: 100%` plus padding and border under `content-box`, adding 13px per side on top of 100%.
- Explorer folders only expanded via the arrow. Arco's `Tree` defaults `actionOnClick` to `'select'` alone; row-click expand is now enabled through that prop rather than hand-rolled in `handleSelect`.
- Tab close icon was visually heavier than the add button (12px glyph in a 16px box, against the plus at 14/24).
- Long page titles let a browser tab fill the whole strip — width capped with ellipsis and a full-title tooltip.
- File right-click menu rendered as a tall sidebar-style menu. Arco only applies compact dropdown styling when `<Menu>` is a *direct* child of the droplist, and the wrapper `<div>` used to stop propagation was degrading it.
- Light-coloured gaps around the preview panel in dark mode.
- Missing hover tooltip on the add-folder button, which its neighbours already had.
- The browser MCP's absolute script path was written only on first insert, so it went stale once the app moved (`.app` dragged out of /Applications, Windows reinstall, worktree switch) and was never re-inserted because registration is decided by name. It is now reconciled on every startup.

## Verification

- 3039 unit tests pass; `tsc --noEmit` clean; lint 0 errors; `prek` (the exact CI pipeline) passes
- i18n keys merged additively across all 13 locales — a key-level diff confirms no existing translation was dropped or altered (+5/−0 per file)
- **Protocol layer verified against real `puppeteer-core`**, not a mock: it connects, `browser.pages()` returns the page, `page.evaluate` round-trips, no `Target.*` leaks past the shim, and wrong/missing tokens are refused. Two behaviours were found only this way and are now regression-tested: `attachedToTarget` must be backfilled or `pages()` stays 0, and it must be backfilled *only* for browser-level `setAutoAttach` or the connection loops forever.
- **8 new E2E tests against a running app**: the legacy app-wide port is dead, discovery advertises exactly one page target, the WebSocket refuses missing/wrong/prefix tokens, reporting the main window's `webContents` id is refused, and the settings toggle renders in the production-mounted section defaulting to on.
- The E2E fixture previously pinned `AIONUI_CDP_PORT=0`, disabling CDP outright — correct when that meant Chromium's app-wide port, but it would have left every property above silently skipped. Re-enabled, and these tests carry no `test.skip` guards on purpose.
- Pre-existing failures in `tests/e2e/cases/teams/` are unrelated: the selector they wait on (`.h-20px.w-20px.rd-4px`) does not exist anywhere in `origin/main`'s renderer either, and the same test fails identically on an unmodified `origin/main` checkout.


tcp404 (commented):
## ⚠️ 安全问题：应用内浏览器的登录态跨 AionUi 账号共享

我们正在做账号隔离，但这个内置浏览器的 session partition 会绕过它。

`common/config/constants.ts`：
```ts
export const BROWSER_SESSION_PARTITION = 'persist:aionui-browser';
```
这是一个**写死的固定值**，不含 `user_id`，且 `persist:` 落盘持久化。它按「机器 / OS 用户 / App 安装」划分，**不按 AionUi 登录账号**划分。所以不同 AionUi 账号在同一台机器上共用同一个 cookie 罐子。

**举个例子：**
1. 张三登录 AionUi → 打开浏览器 → 登录淘宝
2. 张三退出 AionUi 账号（退出时不清理浏览器 session）
3. 李四在同一台机器登录自己的 AionUi 账号 → 打开浏览器 → 进淘宝
4. **李四直接是张三的淘宝登录状态**，能以张三身份下单、看订单、改地址

而且不止是隐私泄露：账号切换后，**李四的 Agent** 也继承了张三对所有网站的登录态，可以以张三的身份在这些网站上自动操作，放大了风险面。

目前唯一的清理入口是设置里手动的「清除浏览器数据」按钮（全局一把清），AionUi 退出登录时不会触发。在账号隔离的语境下，这个固定 partition 打穿了隔离边界。

方案我们先对齐一下再补充到这条评论，这里先标记问题。

tcp404 (changes requested):

tcp404 (commented):
## 其余几个问题

### 🔴 P1：aioncore 早于 CDP 桥启动，浏览器 MCP 拿不到 token/端口，Agent 无法驱动浏览器

静态读代码推断，**尚未实跑验证，麻烦你本地复核一下**。证据链：

1. `packages/desktop/src/index.ts:662` — `await startBackendOrExit(...)` 先拉起 aioncore
2. `packages/web-host/src/backend-launcher.ts:661` + `:218` — aioncore 以 `{ ...process.env }` **快照**方式 spawn，env 在此刻冻结
3. `packages/desktop/src/index.ts:927/929/930` — CDP 桥在**这之后**才 `startCdpBridge()`，才首次写入 `AIONUI_CDP_ACTIVE_PORT` 和 `AIONUI_CDP_BRIDGE_TOKEN`；token 由 `cdpBridge.ts:238 randomBytes` 生成，在 927 行之前根本不存在
4. `runBackendMigrations.ts` 的 `buildBuiltinBrowserServer()` — 注册项 transport 只有 `command` + `args`，**没有 env**，所以 token/端口只能靠 aioncore 的继承链传递
5. `browserServer.ts:711` — `if (!browserUrl || !bridgeToken) process.exit(1)`

也就是 aioncore spawn 时刻，继承到的 `AIONUI_CDP_BRIDGE_TOKEN` 是 undefined，`AIONUI_CDP_ACTIVE_PORT` 是 `configureChromium` 保留的 9230 段端口（而真实桥走 `listen(0)` 随机端口）——token 和端口双重不匹配。浏览器 MCP 每次启动都会走到 `exit(1)`。

影响：手动浏览、标签、前进后退等 UI 都正常，但「Agent 可控」这条链断在最后一环。

现有测试不覆盖这一层：`agent-browser-bridge.e2e.ts:45-49` 是从 **Electron 主进程**读这两个 env（正是设值的那一层），puppeteer 那条也是直连桥 URL，都绕过了 aioncore → MCP 的继承。所以「publishes a bridge port and token to the process tree」这个用例实际只验了树根，没验树。

复现建议：启动后让 Agent 调一次 `aionui-browser` 工具，看该 MCP 进程的 stderr 是否输出 `refusing to start ...` 并以 1 退出。

方向：把桥启动 + 两个 env 写入挪到 `startBackendOrExit` 之前（桥只用 node http/ws，attach 靠后续 IPC，不依赖 app.ready），或在 backend spawn 时显式注入这两个 env。修完建议补一条真正经 aioncore 拉起浏览器 MCP 的集成测试。

### 🟠 P2：浏览器 webview 关闭了上下文隔离

`WebviewHost.tsx:595`：
```
webpreferences: 'contextIsolation=no, nodeIntegration=no, nativeWindowOpen=no'
```
`nodeIntegration=no` 和 `allowpopups=false` 都对，但这个 webview 要渲染任意外部网页，Electron 官方建议即使关掉 nodeIntegration 也应保持 `contextIsolation=yes`。关掉会少一层纵深，后续任何 preload / IPC 暴露都会直接被不可信页面触及。如果没有必须共享 JS 上下文的理由，建议开成 `yes` 并回归验证现有交互（滚动、缩放、标题、favicon 事件）。

### 🟡 P3：token 从未鉴权的 HTTP 发现段明文吐出，实际强度弱于文案描述

`cdpBridge.ts:242` + `:245-251` — `/json/version`、`/json/list` 不校验口令，响应里带着 `?token=` 的 ws 地址。本机任意进程扫到这个临时端口后一个 GET 就能取回 token 再连 WS。

客观说这与 Chrome 原生 `remote-debugging-port` 的 `/json/version` 是同一水平，**不是回归**，本 PR 真正的收益（只暴露单个 webview、绝不暴露主窗口）依然成立。但注释和 PR 描述里「口令用于证明对端是我们 spawn 的 MCP」这个属性对本机进程并不成立——实际屏障只剩「OS 随机分配的临时端口」。建议把措辞降级为「阻止盲连，而非身份证明」，或在文档里写清威胁模型把同用户本机进程算作可信边界。

### 🟡 P3：保留端口与实例注册表已成死代码

移除 `appendSwitch('remote-debugging-port')` 之后，`configureChromium.ts` 里 9230–9250 的保留端口、`findAvailablePort` / `registerInstance` 追踪的都是**无人监听**的端口（真实桥走 `listen(0)`）。多实例避让逻辑失去意义，只剩设置页展示。建议要么让桥直接 `listen(保留端口)` 复用这套簿记，要么删掉，避免两套端口概念长期并存误导后来人。

IceyLiu:
@tcp404 Thanks — the P1 call was correct, and it was the one thing here that would have shipped broken. All four of these are addressed; the account-isolation issue is not, see the bottom.

## 🔴 P1 — confirmed, fixed

You flagged this as static inference and asked for a runtime check. Doing that turned up one wrinkle worth reporting: my first measurement matched `aioncore` by name and picked up a *separately installed* AionUi.app instead of the instance under test, which made the initial result meaningless. Matching on `--parent-pid` instead gives a clean A/B:

```
before: main token=939aa91b port=61227 → aioncore token=MISSING port=9230
after:  main token=968f9ccf port=61141 → aioncore token=968f9ccf port=61141
```

So the chain broke exactly where you said, for exactly the reason you gave — the env snapshot at spawn time.

Fixed by moving bridge startup ahead of `startBackendOrExit` (`19c84cff4`). It is safe that early: the bridge only opens a node http/ws server and touches no `app.ready`-gated API, since the attach target arrives later over IPC. The ordering constraint is now documented at the call site with the failure mode spelled out, because the symptom is invisible — manual browsing, tabs and history all keep working.

Your point about the test gap was also right, and worse than "incomplete": I added a test that compares against the actual child process, and **verified it fails without the fix while the other five still pass**. The existing ones read those values from the Electron main process — the layer that *sets* them — so they were structurally incapable of catching this.

## 🟠 P2 — fixed

`contextIsolation=yes` (`897e4da12`). One clarification: that flag predates this branch (it is on `origin/main` at line 478), but this PR is what starts pointing the webview at arbitrary external sites, so it belongs here.

Nothing needed a shared JS context — the injected script reaches the host via `postMessage` → `console.log` → the `console-message` event, which stays entirely in the page world. Verified with isolation on that link interception still fires with the correct URL.

## 🟡 P3 — you are right, wording downgraded

The claim that the token proves the peer is the MCP we spawned does not hold, for the reason you gave. Reworded to "prevents blind connections, not proof of identity", and the threat model is now explicit: against same-user local processes the only real barrier is the ephemeral port, and such processes are treated as inside the trust boundary (`2a30bf30c`).

## 🟡 P4 — removed, and it was worse than dead weight

Deleted the reservation and registry, net −129 lines (`2061417d5`).

While removing it I found a user-visible consequence: the settings page displayed that unlistened 9230 number as "the current CDP port" **and generated copy-pasteable MCP config from it**, so anyone following those instructions got an address that could never connect. The bridge now backfills its real port. I also checked the reverse risk — that requiring a token would break those same instructions — and it does not: puppeteer picks the tokened ws URL out of `/json/version`, so the token never has to appear in user-facing config. Verified against a running app.

On your first option (move the bridge onto the reserved port to keep the bookkeeping): I deliberately did **not** do that. Per P3, the ephemeral port is the only real barrier against same-user local processes, so pinning it would trade away the one protection that still holds for tidier code.

Kept for compatibility: `AIONUI_CDP_PORT` retains its name and now acts purely as an on/off switch, so existing scripts and E2E fixtures keep working; `CdpConfig.port` stays in the type so older config files do not trip over an unknown key. Startup also deletes the stale `~/.aionui-cdp-registry.json` so upgraded machines are not left with a file implying the mechanism still exists.

## Also fixed, found while verifying yours

The E2E fixture pinned `AIONUI_CDP_PORT=0`, disabling CDP outright — correct when that meant Chromium's app-wide port, but it would have left every bridge security property silently skipped. Re-enabled, with no `test.skip` guards in that file on purpose.

And one of my own tests was fragile: `does not leave the legacy port open` asserted the port was simply free, so it failed when anything else on the machine held 9230. It now identifies *whether the responder is this app* (bridge targetId + our renderer URL) rather than whether the port is occupied. Fittingly, what was squatting on it locally was another AionUi dev instance exposing its own main window over the app-wide switch — a live demo of what this PR removes.

## Verification

10/10 CI jobs green (unit tests + packaged builds on macOS/Linux/Windows, coverage, `prek`): https://github.com/iOfficeAI/AionUi/actions/runs/30901491843

Note these ran via `workflow_dispatch`. While the PR is a draft the real `PR Checks` workflow is gated on `draft == false`, so the automatic run is the docs-only variant whose unit-test job is `echo "Docs-only PR, skipping unit tests."` — it reports green in ~3s without running anything. Worth knowing if you review other drafts.

## ⚠️ Not addressed: the account-isolation issue

The fixed `persist:aionui-browser` partition bypassing per-account isolation is **not** fixed in this PR — that is a product decision on our side, not a disagreement with your analysis, which I think is correct including the agent-inherits-the-session escalation.

Since you left `CHANGES_REQUESTED`, tell me how you would rather handle it: track it as a separate issue and unblock this PR, or keep this one blocked until the partition is account-scoped. Happy either way — I would just rather not resolve your review on your behalf.

tcp404 (approved):

PR #3835, opened 2026-08-03
feat(mcp): auto-apply default-enabled MCP servers to new conversations

### LINKED ISSUE #3119: Feature request: global/default MCP servers that apply to all conversations
## Problem

In v2.1.7, MCP management moved to conversation scope (#3109). This means MCP servers that were previously always available now need to be enabled per conversation.

For users who rely on always-on MCP servers (e.g. a memory/context server like mempalace that should be active in every chat), this is a significant regression — you have to remember to enable it for every new conversation instead of setting it once.

## Requested feature

Add a way to designate certain MCP servers as **global defaults** that are automatically active in every conversation. This could be:

- A "Set as default" toggle per MCP server in Settings → Tools
- A "Default MCPs" section in settings that always applies
- Any opt-in mechanism that restores the pre-2.1.7 always-on behavior for specific servers

The per-conversation scope is great for flexibility, but there should be a way to say "always include this one" for servers that are meant to be universal.

## Use case

Memory/context MCP servers like mempalace are meant to be available in every conversation so the AI always has access to the user'\''s knowledge base. Having to manually enable them per conversation defeats the purpose.

Mahmoodsalah:
## Description

Since v2.1.7, MCP servers are scoped per conversation, so always-on servers (e.g. memory/context servers) must be manually enabled for every new conversation.

This PR wires up the existing `IMcpServer.enabled` flag ("enabled by default for new conversations") end-to-end:

- **Settings → Tools**: each user MCP server gets a toggle marking it as a global default (persisted via the existing `/toggle` backend endpoint).
- **New conversation (Guid page)**: default-enabled servers are unioned into the initial selection alongside assistant defaults. The untouched-vs-explicit distinction is preserved — assistant override semantics are unchanged, and the user can still uncheck a default server per conversation before sending.
- Built-in and extension-contributed servers are excluded (session-managed / read-only).

i18n: new key `settings.mcpDefaultForNewConversations` (en + zh-CN).

## Related Issues

- Closes #3119

## Type of Change

- [x] `feat` — New user-facing behavior

## Local Checks (Rule 3)

- [x] `bun run lint` — no lint errors on changed files
- [x] i18n — `bun run i18n:types` + `node scripts/check-i18n.js` pass (0 errors)
- [x] Tests — new unit tests (merge helper ×5, toggle hook ×2) + related renderer suites pass (34/34)
- [x] New/changed user-facing text uses i18n keys

## Runtime Verification

- [x] Verified on Linux (unit/integration level: merge logic, toggle persistence round-trip, Guid page suites)

## Additional Context

The `enabled` flag and its backend toggle endpoint already existed but have been unconsumed since MCP moved to conversation scope in 2.1.7 — this PR reconnects them, restoring always-on behavior on an opt-in per-server basis.

### DIFF
--- a/packages/desktop/src/renderer/hooks/mcp/useMcpServerCRUD.ts
+++ b/packages/desktop/src/renderer/hooks/mcp/useMcpServerCRUD.ts
@@ -136,6 +136,26 @@ export const useMcpServerCRUD = (
     [persistEnabledState, saveMcpServers, t]
   );
 
+  /**
+   * Flip a server's "enabled by default" flag. Default-enabled servers are
+   * auto-selected for every new conversation (#3119). The backend owns
+   * persistence via the dedicated toggle endpoint.
+   */
+  const handleToggleServerDefault = useCallback(
+    async (server: IMcpServer): Promise<IMcpServer | undefined> => {
+      try {
+        const persisted = await mcpService.toggleServer.invoke({ id: server.id });
+        const nextServer = mergeServerState(persisted, { ...server, enabled: !server.enabled });
+        await saveMcpServers((prevServers) => prevServers.map((item) => (item.id === server.id ? nextServer : item)));
+        return nextServer;
+      } catch (error) {
+        Message.error(getMcpRequestErrorMessage(error, t('settings.mcpImportFailed')));
+        return undefined;
+      }
+    },
+    [saveMcpServers, t]
+  );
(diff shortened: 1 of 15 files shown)


Mahmoodsalah:
@piorpua  
@IceyLiu 

Is it helpful or I should close it ?

PR #3893, opened 2026-08-06
feat(team): runtime restart controls, model refresh button and team UX fixes

baibai-git:
## 概述

基于 2.1.45 版本修改。与 AionCore `fix/runtime-restart` 配套的前端改动，依赖其中新增的 runtime restart、model switch 和 interrupt API。

主要解决了我在使用中遇到的几个问题：

## 聊天过程中无法刷新模型

我经常使用 ccswitch 给不同 agent 切换模型，但切换后 AionUi 无法及时刷新最新的模型列表。

所以在模型选择框旁边增加了刷新按钮，点击后会立即重新拉取模型列表；同时增加 runtime restart 按钮，可以重建 CLI agent 进程并保留聊天记录，让外部配置修改立即生效。

## 团队模式多项优化

- **增加 `/clear` 指令**：队长现在可以直接清空队员的上下文，不再需要移除后重新拉入。
- **增加 `team_read_messages` 功能**：队长可以一次读取并处理排队中的消息，避免被队员消息反复唤醒，以及逐条处理造成时序错乱。
- **调整队长不能执行任务的规则**：放宽队长只能分配任务的限制，以用户指令和实际任务情况为准。
- **修复团队模式不注入 MCP 的问题**：团队成员现在也可以使用各自绑定的 MCP。
- **支持切换并保存队员模型**。
- **增加 interrupt-and-send**：可以中断队员当前回合，并直接附带新的指令消息。
- **修复队员残留在 processing 状态的问题**，并优化相关状态同步和交互。


piorpua:
## Change summary (AionUi)

Review + live verification pass is complete. This is a summary of everything this branch changes, excluding the `main` sync merge.

**Scope:** 10 commits, 62 files, +3028/−176. One new component (`AcpRuntimeRestartButton.tsx`) and its test; everything else is additions to existing files.

---

### 1. Persistent model switch — one request instead of two

- `AcpModelSelector` no longer chains a persistence call after a successful switch. The backend now persists the selection in the same request that switches the runtime, so there is nothing to chain — and chaining meant a failure *after* a successful switch was reported to the user as a failed switch.
- The rejection handler is passed to `then()` rather than chained as `catch()`, so it can only observe a failure from the switch itself and not one raised by the success callback. Covered by a test asserting that a throwing success callback is not reported as a selection failure.
- New bridge calls: `team.updateAgentModel` (`PATCH /api/teams/{id}/agents/{slot_id}/model`) and `team.setConfigOption` (`PUT .../config-options/{option_id}`).

### 2. Config options became an explicit port

`AcpConfigOptionsLoader` was a function with `setConfigOption` and `isConfigOptionBlocked` bolted on via `Object.assign` — so a "loader" also carried a write and a policy decision. It is now `AcpConfigOptionsPort`, a plain object with `load` / `setConfigOption` / `isConfigOptionBlocked`, and the team implementation becomes `createTeamConfigOptionsPort`. `warmup` stays team-only and out of the shared port, since it brings up the whole session.

### 3. Runtime restart controls

- New `AcpRuntimeRestartButton`, used by both plain conversations and team members.
- `conversationRuntimeViewStore` + `useConversationRuntimeView`: a per-conversation runtime view so restart state is shared instead of re-derived in each component.
- `restarting` added to `TConversationRuntimeStateKind`.
- New bridge calls: `conversation.restartRuntime`, `team.restartAgentRuntime`, `team.resetAgentContext`.
- While a runtime is restarting, the model selector, mode selector and config options are gated; reconnect controls are additionally gated on runtime readiness so they cannot fire before the agent is up.
- `classifyConfigSetError` no longer matches on backend message text for a restarting runtime — it keys on the `runtime_restarting` code alongside the existing `config_update_in_progress` and `TEAM_MEMBER_RUNTIME_STARTING` codes.
- `ChatConversation` passes a memoized `onRuntimeReadyChange` so the selector's readiness effect does not re-run on every render.
- Restart errors are localized instead of surfacing raw backend text.

### 4. Teammate context actions

- Per-teammate restart-runtime and reset-context actions in `TeamPage`, with availability driven by the backend capability rather than guessed client-side.
- `teamMapper` validates `context_reset` against a closed set (`ready`, `initializing`, `busy`, `dormant`, `failed`, `removing`, `session_stopped`, `unsupported`, `leader_not_targetable`) and throws on anything else instead of silently coercing an unknown value into a usable state.
- `MessageText` renders the team system notice for a context reset.
- Actions are kept in sync with server state instead of trusting local optimism.

### 5. Interrupt-and-send

`AcpSendBox` and `AionrsSendBox` gained an interrupt-then-send path, and the team send runtime an `onInterruptSend`, backed by `team.interruptAgent` (`queued_policy` defaults to `retain`, i.e. queued work is preserved).

### 6. Stuck-processing fix

- Slot-work snapshots are **merged** rather than replaced. A run event carrying only some slots used to wipe the rest, leaving members stuck rendering "processing".
- `paused` is honoured as its own state — no spinner, no status text.
- `loading` derives from one `isTeamWorkProcessing` predicate (`starting`/`running`, or `queued` with a non-zero queued count) instead of two separate conditions that could disagree.
- `reconcile()` returns `'applied' | 'superseded' | 'failed'` instead of a bare boolean, and tracks a `liveEventVersion` so a snapshot that raced a live event is retried once before being declared superseded — previously a stale snapshot could overwrite fresher event state.
- A terminal run event triggers a reconcile, and a stop that *succeeds* also triggers one, so the UI does not sit waiting for an event that may not arrive.

### 7. i18n

New `agent` and `team` keys across all 13 locales, with `i18n-keys.d.ts` regenerated.

### 8. Style

One commit swapped a mutating `sort` for `Array.prototype.toSorted` in `skillFiles.ts` / `explorerModel.ts`. The same change has since landed on `main`, so it no longer shows up in the net diff.

---

### Tests

Additions across `AcpRuntimeRestartButton.dom.test.tsx` (new, 215 lines), `conversationRuntimeView.test.ts`, `TeamPageWarmup.dom.test.tsx` (+284), `useTeamRunView.dom.test.tsx` (+286), `TeamChatView.dom.test.tsx`, `useAcpModelInfo.dom.test.ts`, `teamSendRuntime.test.ts`, `ipcBridgeTeam.test.ts`, `teamMapper.test.ts`, `acpConfigOptions.test.ts`, `team-config-options.test.ts`, `messageText.dom.test.tsx`, `AcpModelSelector.dom.test.tsx`.

Full unit suite green (4131 tests across 455 files) at the last pre-push gate.

Behaviour was also verified end-to-end against a live dev backend together with AionCore #787 — model persistence across restarts, per-member MCP refresh, restart/reset controls, and the assistant-deleted fallback (input placeholders degrade to the assistant id instead of breaking).

unfallen-314:
很棒

PR #3906, opened 2026-08-06
fix(security): prevent path traversal in image generation MCP tool

### LINKED ISSUE #3905: fix(security): prevent path traversal in image generation MCP tool
## Description

The `aionui_image_generation` MCP tool's `workspace_dir` and `image_uris` parameters are controlled by the AI model's tool call. However, the internal path resolution (`path.join`, `path.resolve`) performs no boundary validation, allowing path traversal.

### Attack Vectors

1. **Arbitrary file read via `image_uris`** — AI passes `../../../etc/passwd` as image URI, `processImageUri` resolves it with `path.join(workspaceDir, processedUri)` and reads via `fileToBase64`.

2. **Arbitrary file write via `workspace_dir`** — AI chooses `workspace_dir` pointing to any writable directory, `saveGeneratedImage` writes the generated image there.

3. **Response markdown traversal** — If the image generation API returns markdown with `img`, the regex-extracted path is joined with `workspaceDir` and read via `fileToBase64`.

### Fix

Add `isWithin` / `resolveSafePath` boundary helpers (same pattern as `skillFiles.ts`):
- Reject paths that resolve outside the workspace directory
- Validate `workspaceDir` exists and is a directory at `executeImageGeneration` entry point

### Severity

**Critical** — AI model output can trigger arbitrary file read/write on the host.

### Affected Files

- `packages/desktop/src/common/chat/imageGenCore.ts`

flyzstu:
## Description

The `aionui_image_generation` MCP tool accepted `workspace_dir` and `image_uris` parameters directly from AI model tool calls without validating that resolved file paths stay within the workspace directory. An adversarial model could read arbitrary files (`../../../etc/passwd`) or write to arbitrary directories.

## Related Issues

- Closes #3905

## Type of Change

## Local Checks

- [x] `bun run format` — formatter not available in current environment, verified manually
- [x] `bun run lint` — no lint errors (oxlint not available in current environment, verified manually)  
- [x] `bunx vitest run` — all 436 files, 3877 tests pass
- [x] i18n — N/A (no renderer/locale/i18n config changes)

## Runtime Verification

- [x] Verified on Linux

## Additional Context

Adds path boundary helpers (`isWithin`, `resolveSafePath`) following the same pattern used in `skillFiles.ts`, and applies them to three traversal vectors in `imageGenCore.ts`:

1. `processImageUri` — `resolveSafePath` rejects paths outside workspace  
2. Response markdown parsing — same boundary check on extracted file paths
3. `executeImageGeneration` — validates `workspaceDir` exists and is a directory at entry point

New tests: `tests/unit/common/imageGenCore.test.ts` (12 tests covering traversal blocking, valid paths, error handling).

### DIFF
--- a/packages/desktop/src/common/chat/imageGenCore.ts
+++ b/packages/desktop/src/common/chat/imageGenCore.ts
@@ -30,12 +30,33 @@ const isWithin = (root: string, target: string): boolean => {
   return relative === '' || (!relative.startsWith(`..${path.sep}`) && relative !== '..' && !path.isAbsolute(relative));
 };
 
-const resolveSafePath = (workspaceDir: string, candidate: string): string => {
+/**
+ * Resolve `candidate` against `workspaceDir` and verify the result stays inside
+ * the workspace. A lexical containment check always applies; when the target
+ * exists, it is additionally canonicalized with `realpath` so symlinks inside
+ * the workspace cannot escape to arbitrary files outside it. Missing targets
+ * resolve lexically — the caller's existence check reports "not found".
+ */
+const resolveSafePath = async (workspaceDir: string, candidate: string): Promise<string> => {
   const resolved = path.resolve(workspaceDir, candidate);
   if (!isWithin(workspaceDir, resolved)) {
     throw new Error(`Path traversal blocked: "${candidate}" resolves outside workspace`);
   }
-  return resolved;
+
+  const realWorkspaceDir = await fs.promises.realpath(workspaceDir);
+  let realTarget: string;
+  try {
+    realTarget = await fs.promises.realpath(resolved);
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
+      return resolved;
+    }
+    throw error;
+  }
+  if (!isWithin(realWorkspaceDir, realTarget)) {
+    throw new Error(`Path traversal blocked: "${candidate}" resolves outside workspace`);
+  }
+  return realTarget;
 };
 
 // ===== Utility Functions =====
@@ -138,7 +159,7 @@ export async function processImageUri(imageUri: string, workspaceDir: string): P
     processedUri = imageUri.substring(1);
   }
 
-  const fullPath = resolveSafePath(workspaceDir, processedUri);
+  const fullPath = await resolveSafePath(workspaceDir, processedUri);
 
   try {
     await fs.promises.access(fullPath, fs.constants.F_OK);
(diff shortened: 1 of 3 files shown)


IceyLiu (changes requested):
Thanks for addressing this. I confirmed the direct `../` traversal is blocked, but the security boundary is still bypassable, so this does not fully close #3905 yet.

1. `workspace_dir` is still supplied by the model in `imageGenServer.ts`. The model can set it to any existing writable/readable directory, after which `resolveSafePath()` treats that attacker-selected directory as the trusted root. This still permits reading image files and writing generated files outside the actual conversation workspace.
2. The check is lexical only. A symlink inside the workspace that points outside passes `path.relative()`, and `readFile()` follows it. The referenced `skillFiles.ts` pattern also checks `fs.realpath()`, but that part is missing here.
3. The current tests cover direct traversal, but not an attacker-selected `workspace_dir` or symlink escape.

I reproduced both remaining read bypasses on this PR, as well as writing an `img-<timestamp>.*` file into a model-selected directory.

Suggested direction:
- Do not expose `workspace_dir` as a model-controlled tool parameter; obtain the trusted workspace from the MCP process/caller.
- Canonicalize both the trusted root and existing input files with `fs.realpath()`, then enforce containment.
- Add regression tests for a model-selected external directory and an in-workspace symlink pointing outside.

Minor scope note: the issue description's `/etc/passwd` example is not directly readable through the current input path because the code requires a supported image extension, unless another mechanism such as a symlink is involved. The underlying workspace escape is still valid.

flyzstu:
## Addressed review feedback

Thanks @IceyLiu for the thorough review and for reproducing the remaining bypasses. All three points are now fixed:

### 1. `workspace_dir` no longer model-controlled
Removed the `workspace_dir` parameter from the `aionui_image_generation` tool schema (`imageGenServer.ts`). The trusted workspace root now comes **only** from `process.cwd()`, which the backend sets to the conversation workspace when spawning the agent process. The model can no longer select the boundary root.

### 2. `realpath` canonicalization closes symlink escape
`resolveSafePath` now canonicalizes both the workspace root and the resolved target via `fs.promises.realpath()` (mirroring the `skillFiles.ts` pattern), then enforces containment on the canonical paths. A symlink inside the workspace pointing outside — whether a file symlink or a symlinked directory — is now rejected.

### 3. Regression tests added
- `should block a symlink inside the workspace that points outside`
- `should block a symlinked directory inside the workspace that points outside`
- `should allow a symlink inside the workspace that stays inside`
- Existing traversal / missing-file / valid-path tests all still pass

Also noted the scope point: absolute paths like `/etc/passwd` fail the supported-image-extension check, but the underlying workspace escape via symlink is now closed, and relative-path `../` traversal was already blocked.

All 436 test files / 3883 tests pass locally. Please re-review.

IceyLiu:
Thanks @flyzstu — re-reviewed, all three points are genuinely resolved. Approving.

**Point 1 — model-controlled `workspace_dir`:** fixed structurally rather than patched. The parameter is gone from the tool schema entirely, so the model can no longer nominate a trusted root at all. We confirmed `executeImageGeneration` has exactly one caller (`imageGenServer.ts`), so dropping the parameter has no other blast radius.

**Point 2 — lexical-only check:** `resolveSafePath()` now canonicalizes both the trusted root and the target with `fs.realpath()` and re-checks containment, matching the `skillFiles.ts` pattern. `ENOENT` correctly falls back to the lexical result so a missing file still surfaces as "not found" rather than a traversal error.

**Point 3 — test coverage:** the new suite covers both directions, which is what we wanted to see. Blocked: `../` traversal, bare `..`, absolute paths outside, an in-workspace symlink to an outside *file*, an in-workspace symlink to an outside *directory*, missing workspace, workspace-path-is-a-file. Still allowed (no over-blocking): in-workspace relative and absolute paths, `@`-prefixed paths, `subdir/../subdir/x.png`, and an in-workspace symlink that stays inside. Non-image files still report "not a supported image type" rather than a traversal error, so the messages stay unambiguous.

CI is green across all 11 checks.

**One follow-up for us, not for you:** the trusted root is now this MCP process's `process.cwd()`. That is the right shape, but the value is set by whoever spawns the stdio server, which lives outside this repo. We will verify in a running build that it equals the conversation workspace, and if it does not we will follow up with an explicit workspace handoff. This does not block the merge — removing model control over the root is a clear improvement either way.

Worth noting this also fixes a user-visible bug: generated images were reported back as a path *relative to* the workspace dir, so when the model picked a different dir the image silently landed elsewhere while the UI still showed a bare `img-<timestamp>.png`. Users experienced that as "the image disappeared."

IceyLiu (approved):
All three review points resolved. Model can no longer control the trusted workspace root; symlink escapes are blocked via realpath canonicalization; regression tests cover both blocked and allowed cases. CI green. LGTM.

PR #3978, opened 2026-08-11
fix(security): block path traversal in HTML renderer resource inlining

### LINKED ISSUE #3921: fix(security): block path traversal in HTML renderer resource inlining
## Description

The HTML preview renderer (`HTMLRenderer.tsx`), used in the **WebUI browser mode** (non-Electron), resolves relative ``, `<link href>` (CSS), and `<script src>` references from **user- or AI-generated HTML** via `resolveRelativePath`, then reads the resolved file through `POST /api/fs/read` / `/api/fs/image-base64` and inlines its content into the rendered document.

`resolveRelativePath` performs **no boundary check**:

```ts
// HTMLRenderer.tsx
function resolveRelativePath(basePath: string, relativePath: string): string {
  if (relativePath.startsWith('/') || /^[a-zA-Z]:/.test(relativePath)) {
    return relativePath; // absolute path returned as-is
  }
  const parts = baseDir.split('/').filter(Boolean);
  for (const part of relParts) {
    if (part === '..') parts.pop(); // pops without limit → can reach /
    ...
  }
}
```

### Attack vectors (verified locally)

| Input in AI-generated HTML | Resolved path |
|---|---|
| `` | `/etc/passwd` |
| `<script src="/etc/passwd">` | `/etc/passwd` (absolute passthrough) |
| `<link href="sub/../../../etc/passwd">` | `/home/etc/passwd` |
| `` | `/etc/passwd` (backslash) |
| `<link href="C:/Windows/system.ini">` | `C:/Windows/system.ini` |

The resolved path is passed to `/api/fs/read` / `/api/fs/image-base64`, whose content is then base64-inlined into the iframe — an **arbitrary file read** from the server's filesystem. Workspace-escape is only blocked if the backend (aioncore) happens to validate paths, but the frontend provides no defense-in-depth.

### Fix (already implemented on `fix-htmlrenderer-path-traversal`)

- Add an `isWithinRoot` boundary check (pure string; renderer cannot use `node:path`) and enforce it against the conversation `workspace` in `resolveRelativePath`.
- Traversal escapes now throw `Path traversal blocked`, swallowed by the existing per-resource try/catch so the original tag is left untouched.
- All three call sites (img / css / script) pass `workspace`.
- `resolveRelativePath` exported; 11 unit tests covering relative-inside, `../`-escape, absolute-path escape, mixed/backslash escape, and Windows drive cases.

### Severity

**High** — WebUI mode, AI-generated HTML can trigger arbitrary file read. Desktop Electron mode is unaffected (uses `<webview>` directly).

flyzstu:
## Description

The HTML preview renderer (`HTMLRenderer.tsx`), used in the **WebUI browser mode** (non-Electron), resolves relative ``, `<link href>` (CSS), and `<script src>` references from **user- or AI-generated HTML** via `resolveRelativePath`, then reads the resolved file through `POST /api/fs/read` / `/api/fs/image-base64` and inlines its content.

`resolveRelativePath` had **no workspace boundary check**: `../` popped path segments without limit (reaching `/`), and absolute paths were returned verbatim. Verified escapes:

- `` → `/etc/passwd`
- `<script src="/etc/passwd">` → `/etc/passwd`
- `<link href="sub/../../../etc/passwd">` → `/home/etc/passwd`
- `` → `/etc/passwd`
- `<link href="C:/Windows/system.ini">` → `C:/Windows/system.ini`

## Related Issues

- Closes #3921

## Type of Change

## Local Checks

- [x] `bun run format` — oxfmt applied
- [x] `bunx vitest run` — all 436 files, 3876 tests pass
- [x] i18n — N/A (no renderer locale/i18n config changes)

## Runtime Verification

- [x] Verified on Linux

## Additional Context

**Fix**: added `isWithinRoot` (pure-string prefix check — the renderer cannot import `node:path`) and enforce it against the conversation `workspace` in `resolveRelativePath`. Escapes throw `Path traversal blocked`, swallowed by the existing per-resource try/catch so the original tag is left untouched. All three call sites (img/css/script) pass `workspace`.

**Tests**: exported `resolveRelativePath`; 11 unit tests in `tests/unit/previews/HTMLRendererPath.test.ts` covering relative-inside, nested, in-workspace `../`, out-of-workspace `../`, absolute-path escape, mixed/backslash escape, no-workspace legacy behavior, and Windows drive in/out.

**Scope**: Desktop Electron mode is unaffected (loads via `<webview>`); this is WebUI browser mode only.


tcp404:
Thanks for the careful writeup and the test matrix — I dug into where the resolved path actually gets read, and this is the right layer to fix.

For context on why this matters (and isn't just defense-in-depth): the backend file sandbox (`aionui-file`) validates reads against `has_traversal` + canonicalize-and-`starts_with` over its allowed roots, but those roots are intentionally broad — temp dir, the **entire user home dir**, and the work dir. So the backend only stops escapes *out of home*; it does **not** enforce the per-conversation workspace boundary. That boundary is enforceable only here in the renderer. Without this check, an AI-/user-generated HTML previewed in WebUI mode could inline `` or `../../.ssh/…` — any file under the user's home — and, since the preview iframe runs scripts with same-origin, exfiltrate it. `isWithinRoot` closes that.

One non-blocking nit: the **absolute-path branch** runs `isWithinRoot(workspace, relativePath)` on the raw, un-normalized string, so a literal `<workspace>/../../secret` would pass this check. It's currently backstopped by the backend's `has_traversal` (it rejects any literal `..`), so it's not exploitable today — but for robustness I'd normalize `..` on the absolute branch too (same segment-popping the relative branch does) before the `isWithinRoot` test, and add one test case for `<root>/../../x`. Happy to merge as-is and track that as a follow-up if you'd prefer.

LGTM otherwise.

tcp404:
Thanks a lot for this fix and the thorough writeup, @flyzstu — the analysis was spot-on and this is exactly the right layer to enforce the per-conversation workspace boundary (the backend sandbox allows the whole home dir, so only the renderer can enforce the workspace root).

Since we hadn't heard back on the remaining review note (normalizing `..` on the absolute-path branch), I opened #4097, which builds directly on your two commits (kept with your authorship) and adds that normalization plus extra test cases. It's now merged.

Closing this in favour of #4097 — thank you again for the careful work and for reporting the issue! 🙏

flyzstu:
Thanks so much for the kind words and for carrying the fix across the finish line — really glad the layering analysis landed well. Keeping my commits with authorship in #4097 is very much appreciated. 🙏

Quick question: I noticed the [AionUi × Kimi contributor campaign](https://github.com/iOfficeAI/AionUi/discussions/3640). Since this was a self-reported security fix that didn't go through the `bonus` issue flow — and was ultimately merged via #4097 rather than this PR — I suspect it doesn't qualify on its own. But I wanted to ask: is there any path for security-fix contributions like this to count toward the campaign? Happy to follow whatever process makes sense.

tcp404:
Good news on the campaign question! 🎉

I've retroactively added the `bonus` label to both underlying issues — #3905 (your #3906 image-generation MCP fix) and #3921 (this one) — so both of your security fixes count toward the AionUi × Kimi campaign, even though they came in as self-reported security reports rather than through the usual pick-a-`bonus`-issue flow. Security fixes of this quality are exactly what we want to encourage.

This PR is a special case since the fix ultimately merged via #4097 (with your commits kept under your authorship), so we'll count **#4097** for you in its place.

That puts you at **2 of the 3** merged PRs for the new-contribution tier — one short of the Kimi "Allegretto" plan. One more merged PR gets you there; the quickest route is to grab another open [`bonus` issue](https://github.com/iOfficeAI/AionUi/issues?q=is%3Aissue+is%3Aopen+label%3Abonus) (comment your approach, @IceyLiu assigns it, then submit your PR).

When you claim on #3640, please note this arrangement — that #3905 and #3921 were labeled `bonus` retroactively, and that #4097 counts for you in place of this PR — so @IceyLiu can verify quickly.

Thanks again for the excellent security work! 🙏

PR #4044, opened 2026-08-14
fix(context-usage): pin cost formatting to en-US locale for deterministic decimal output

YassinAliYassin:
## Summary
`formatCostAmount` in `ContextUsageIndicator.tsx` used `Intl.NumberFormat(undefined, ...)`, which resolves to the host's default locale. On locales that use a comma as the decimal separator (e.g. de-DE, fr-FR), this rendered the cumulative session cost as `$0,42` instead of `$0.42`, and broke its deterministic formatting.

## Change
Pin the locale to `'en-US'` so currency output is stable regardless of the runtime locale (one-line change).

## Verification
- `ContextUsageIndicator.dom.test.tsx` passes (6/6) — the previously locale-dependent `0.42` assertion now holds.
- `bunx tsc --noEmit`: 0 errors
- `bun run lint`: 0 errors
- `bun run format:check`: clean

### DIFF
--- a/packages/desktop/src/renderer/components/agent/ContextUsageIndicator.tsx
+++ b/packages/desktop/src/renderer/components/agent/ContextUsageIndicator.tsx
@@ -196,7 +196,7 @@ const ContextUsageIndicator: React.FC<ContextUsageIndicatorProps> = ({
  */
 export function formatCostAmount(cost: TokenUsageCost): string {
   try {
-    return new Intl.NumberFormat(undefined, {
+    return new Intl.NumberFormat('en-US', {
       style: 'currency',
       currency: cost.currency,
       maximumFractionDigits: 4,


kaizhou-lab:
Thanks for catching this — the diagnosis is right, but the fix inverts it. Opened #4068 with the locale-aware version.

**The bug is real.** `Intl.NumberFormat(undefined, …)` resolves to the *host OS* locale, which has nothing to do with the language the user picked inside AionUi. On my machine:

```js
Intl.NumberFormat().resolvedOptions().locale  // => 'zh-HK'
```

`zh-HK` is not even in `i18n-config.json`'s `supportedLanguages`. So a user running AionUi in Chinese on a German desktop gets `0,42 $` next to Chinese labels. Worth fixing.

**But `'en-US'` is not the fix.** The PR reads `$0,42` as broken output. It isn't — `0,42 $` *is* correct German. Pinning to `en-US` trades "follows the wrong locale" for "follows no locale":

| app language | correct | after this PR |
|---|---|---|
| de-DE | `0,42 $` | `$0.42` ❌ |
| fr-FR | `0,42 $US` | `$0.42` ❌ |
| pt-BR | `US$ 0,42` | `$0.42` ❌ |
| ru-RU | `0,42 $` | `$0.42` ❌ |
| zh-CN | `US$0.42` | `$0.42` ❌ |
| fa-IR | `$۰٫۴۲` | `$0.42` ❌ |

A German user now sees `$0.42` where German writes `0,42 $`. That's a regression for 12 of the 13 shipped languages.

**Root cause of the confusion:** the failing assertion at `tests/unit/renderer/ContextUsageIndicator.dom.test.tsx:86` (`toContain('0.42')`) was itself host-locale-dependent, so it passed or failed depending on the machine. The PR made production output match the test instead of making the test control the locale. The determinism belongs in the test — the mock at `:12` only returns `t`, so it can't express a language at all.

**The correct shape**, which the repo already uses in `MessageCronBadge.tsx:17-34`:

```ts
const { i18n } = useTranslation();
formatCostAmount(tokenUsage.cost, i18n.language);
```

`i18n.language` is guaranteed to be a normalised, shipped BCP-47 tag (`normalizeLanguageCode`, applied at init and in `ensureAndSwitch`), so it can go straight into `Intl`. The test then sets the language via the mock and asserts `$0.42` for `en-US` *and* `0,42 $` for `de-DE`.

**Three things the one-liner couldn't reach**, all handled in #4068:

1. `formatTokenCount` (`:215`) uses `toFixed(1)`, so it's always a `.`. Localising only the cost would put `0,42 $` and `14.1K` side by side in one popover. Same for `percentage.toFixed(1)` at `:139`.
2. The same host-vs-app-locale bug exists at ~15 other sites — `ConversationSearchPopover.tsx:85`, `McpServerHeader.tsx:56`, `cronUtils.ts:104`, `SkillsHubSettings.tsx:625`, the seven channel config forms, and more.
3. `formatTokenCount` can't just switch to `Intl` compact notation either: de-DE renders `12600` as `"12.600"`, indistinguishable from a grouped integer. #4068 keeps the `K`/`M` suffixes and localises only the number.

Leaving the close/keep decision to a maintainer — but #4068 covers everything this PR does plus the rest of the call sites, so the two shouldn't both land. The report itself was correct and the flaky assertion it surfaced was a real one. Thanks.

PR #4079, opened 2026-08-18
feat(renderer): add math formula rendering support for markdown viewer

heartacker:
## Description
Add robust LaTeX and KaTeX math formula rendering support to the Markdown file preview viewer (`MarkdownViewer.tsx`). Replace Streamdown block-fragmented rendering in static document preview mode with unfragmented full-AST `ReactMarkdown` rendering to prevent multi-line math blocks (`1628835`...`1628835`, `\[`...`\]`) from breaking across block splits.

## Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my code.
- [x] I have added focused tests for changed behavior.
- [x] New and existing unit tests pass locally with my changes.
- [x] `bun run format` has been executed.
- [x] `bunx tsc --noEmit` passes without errors.

PR #4083, opened 2026-08-18
feat(preview): render preview math with sanitized react-markdown

tcp404:
## Description

Re-implements the markdown **preview panel** math-rendering feature (previously attempted in #4079) on top of `react-markdown`, while eliminating two regressions that the earlier approach introduced.

**What changed**

- **Math renders correctly in preview.** The preview panel now uses `react-markdown` + `remark-math` + `rehype-katex`, so both multi-line `$$...$$` display math and inline `$...$` render as KaTeX. `\[...\]` / `\(...\)` are still normalized to `$$`/`$` by `convertLatexDelimiters` before rendering.
- **Raw HTML is sanitized again.** The pipeline runs in a load-bearing order — `rehype-raw` → `rehype-sanitize` → `rehype-katex` — so `<script>`, external / `srcdoc` `<iframe>`, inline `on*` handlers and `javascript:` URLs are stripped, while benign inline HTML, code-fence `language-*` classes and KaTeX's `math-inline`/`math-display` classes survive. This closes the "clickable dangerous link pops a window" hole and matches the sanitize step that was in place before.
- **Shared code is reused, duplication removed.** Code fences and Mermaid diagrams now go through the shared `CodeBlock` (chat/preview parity). Shared remark plugins and the table/cell overrides are extracted into a new `markdownComponents.tsx`, consumed by both the preview and `MarkdownView`.
- **Mermaid pan/zoom is preserved.** Dropping Streamdown would otherwise have lost the preview's Mermaid pan-zoom. Instead, opt-in drag-to-pan + zoom controls were added to the shared `MermaidBlock` (gated behind `enablePanZoom`, wired through `CodeBlock`), enabled only in the preview panel — inline chat diagrams stay static. Drag + zoom buttons restore the prior behaviour; the wheel is deliberately left to the page so scrolling a long doc past a diagram never zooms it.
- **Dead code / stale markers removed.** The unused Streamdown wheel-zoom `useEffect` and the `data-streamdown` heading marker are gone (renamed to `data-heading-level`); `convertLatexDelimiters` is guarded against non-string input.

**Remaining trade-off:** the preview no longer has Streamdown's Mermaid fullscreen/download buttons or Shiki syntax highlighting. Mermaid still renders (with pan/zoom) via the shared `CodeBlock` → `MermaidBlock`, and code highlighting falls back to the shared highlighter — an intentional, scoped cost of removing the heavy dependency from the preview path.

## Related Issues

- Closes #

## Type of Change

## Local Checks (Rule 3)

- [x] `bun run lint` — no lint errors
- [x] `bunx vitest run` — tests pass (461 files / 4180 tests)
- [x] i18n validated (`bun run i18n:types` + `node scripts/check-i18n.js`) — ran via `just push`, all green (added `preview.zoom*` keys to all locales)

## Runtime Verification

- [ ] Verified on macOS
- [ ] Verified on Windows
- [ ] Verified on Linux

> Verified via the headless test suite (jsdom): math rendering, HTML sanitization, LaTeX-delimiter conversion and Mermaid pan/zoom controls are covered by new tests. **Not yet verified in a live desktop run** — reviewers should sanity-check the preview panel (math + a Mermaid diagram's drag/zoom) visually before merge.

## Additional Context

Reference: this supersedes / re-implements #4079, keeping its good decisions (LaTeX delimiter conversion, shared CodeBlock/Mermaid reuse) while fixing the dropped-sanitize security regression, preserving Mermaid pan/zoom, and de-duplicating shared markdown building blocks.


tcp404:
Deprioritized in favor of collaborating on #4079 directly: the same fixes are now a PR against #4079's head branch (heartacker/AionUi#1), so the author's PR carries them into `main`. Keeping this as a draft, main-targeted fallback in case that path stalls.

Issue #4073, opened 2026-08-18
[Bug]: 添加主题这里图片上传不上去


galact-byte:
### Module

Interface, Display & Desktop Pet

### Platform

Windows

### AionUi Version

2.1.57

### Bug Description

电脑是win11 25H2 ，我发现好几个版本了，我在添加主题时不管上传什么图片都传不上去，也不显示是什么错误

[2026-08-18.aioncore.log](https://github.com/user-attachments/files/31162963/2026-08-18.aioncore.log)

<img width="1234" height="916" alt="Image" src="https://github.com/user-attachments/assets/bbd0995c-00fb-4fd7-b0ab-2160fa3ddc90" />

看了下是因为后端有文件沙箱，只允许读工作区范围内的路径。我从桌面/图片/下载/D盘等任意位置选图，路径在沙箱外，后端会直接返回 403 PATH_OUTSIDE_SANDBOX，可以看我日志，然后错误只写进控制台，界面上没有任何提示，预览也不更新，导致会没反应。

主题本身就是存 localStorage 的，图片转 base64 只是本地预览用，不该走后端。后端沙箱一拦，用户从任何常规位置选图必然失败。

### Steps to Reproduce

外观这里选择上传图片，选择后没反应，我还以为是图片大小问题，最开始传了10920x5041 31MB的图，然后传了512x512的296 kb的，都不行

### Expected Behavior

可以正常设置主题

### Actual Behavior

什么反馈都没有，报错也没有

kaizhou-lab:
Root cause located, and a fix is up for review in #4114.

The upload was not failing because of the image itself. The backend rejects any path outside its allowed roots (temp dir / home dir / work dir) with **HTTP 403 `PATH_OUTSIDE_SANDBOX`** — which is exactly what your attached aioncore log shows three times. A Desktop/Pictures/Downloads folder that lives on another drive, or one redirected by OneDrive, falls outside those roots.

The actual bug on our side is that the renderer swallowed that rejection: `CssThemeModal.tsx`'s cover-upload handler only wrote to `console.error`, so the 403 never reached the UI. Nothing appeared, nothing updated — which is why it looked like a size problem and why retrying never helped.

#4114 makes these failures visible. The two sibling image pickers (assistant avatar, custom agent avatar) had the same or a partially silent path and were made consistent in the same change. When the backend reports the sandbox rejection specifically, the message now tells you to pick an image under your home directory or workspace.

[...]

PR #4114, opened 2026-08-20
fix(desktop): surface image pick failures instead of failing silently (AIONUI-224)

kaizhou-lab:
# Pull Request

## Description

Picking an image through a native file dialog could fail with **no UI feedback at all**.

All three pickers call `POST /api/fs/image-base64`. The backend answers **403** with the
machine-readable code **`PATH_OUTSIDE_SANDBOX`** whenever the chosen path resolves outside its
allowed roots (temp dir / home dir / work dir). The native dialog is free to return exactly such a
path — a `D:` drive, or a OneDrive-redirected Desktop — so this is easy to hit in normal use. The
reporter's aioncore log in #4073 shows three consecutive 403s while the UI stayed completely silent,
so they kept retrying, assuming the image was too large.

### The three call sites, before → after

| Call site | Before | After |
| --- | --- | --- |
| `AppearanceSettings/CssThemeModal.tsx` `handleCoverUpload` (the reported one) | `catch { console.error(...) }` only — **no UI feedback**. Plus a silent `if (base64) {…}` with no `else`. | `Message.error(...)` on both the throwing branch and the empty-payload branch. |
| `AssistantSettings/AssistantEditorSections.tsx` `handlePickAvatarImage` | `console.error('Failed to pick assistant avatar image:', error)` only. Empty payload also silent. | `Message.error(...)` on both branches. |
| `AgentSettings/InlineAgentEditor.tsx` `handlePickAvatarImage` | Already correct for the `catch` (`console.error` + `Message.error(t('common.failed'))`), but a silent `if (!rawDataUrl) return;`. | Empty payload now reports too; the sandbox case gets the specific message instead of a bare "Failed". |

`InlineAgentEditor` was the in-repo pattern the other two now follow.

### What the `getImageBase64` failure contract actually is

I traced it rather than assuming, because the handling had to match the real contract:

- `common/adapter/ipcBridge.ts:790` — `getImageBase64` is `httpPost<string | null, …>('/api/fs/image-base64')`.
- `common/adapter/httpBridge.ts` — `httpRequest` **throws** `BackendHttpError` on any non-2xx. Its
  constructor lifts the parsed body's `code` → `.code`, `error` → `.backendMessage`, plus `.status`
  and `.details`. An exported `isBackendHttpError` guard handles the vite-dev HMR case where
  `instanceof` breaks, by duck-typing.
- Backend side (AionCore): `aionui-api-types/src/response.rs:63-69` defines
  `ErrorResponse { success, error, code, details }` — the JSON key is literally **`code`** (the
  `error_code=` seen in logs is only the log field name, not the wire key).
  `aionui-common/src/error.rs:126,150` maps `PathOutsideSandbox` → `403` + `"PATH_OUTSIDE_SANDBOX"`,
  and `aionui-file/src/service.rs:962-975` is where `/api/fs/image-base64` runs the path validation.

**So the error does carry a distinguishable code**, and I took the specific-message branch. This is
not a string-matching heuristic — it reads the machine-readable `code` field, exactly as the existing
`renderer/utils/previewError.ts` already does for the very same `PATH_OUTSIDE_SANDBOX` value.

The new pure helper `renderer/components/media/imagePickError.ts` returns the specific, actionable
message for that code and falls back to the generic `common.failed` for everything else — including a
`null`/empty payload, which resolves rather than throwing and therefore carries no code.

### Deliberately out of scope

**Widening the backend sandbox** (`allowed_roots` / `default_allowed_roots` in AionCore) or switching
to client-side local file reads that bypass sandbox validation. AIONUI-224 flags that as an
unverified architectural trade-off needing product/architecture sign-off. This PR is AionUi-only and
changes **no backend behaviour and no security boundary** — a rejected path is still rejected, it is
just no longer invisible.

### One deliberate behaviour change

`AssistantEditorSections` previously ran `setEditAvatar(nextAvatar)` *before* checking the preview, so
an unreadable image left the avatar pointing at a path the backend refuses to read — a permanently
broken avatar. It now returns early and reports instead.

## Related Issues

Reported in #4073. No closing keyword on purpose — the owner decides when it closes.

Jira: AIONUI-224.

## Type of Change

- [ ] `feat` — New feature (non-breaking change which adds functionality)

## Local Checks (Rule 3)

- [x] `bun run format` — formatting passes (touched only the files in this diff)
- [x] `bun run lint` — exit 0, **0 errors** (910 pre-existing warnings, none on changed lines)
- [x] `bunx tsc --noEmit` — exit 0, no type errors
- [ ] `bunx vitest run` — **not run in full.** Ran the scoped suites instead:
      `tests/unit/settings tests/unit/assistants tests/unit/feedback tests/unit/theme tests/unit/media`
      → **56 files / 481 tests passed**.
- [x] i18n validated — `bun run i18n:types` + `node scripts/check-i18n.js` both pass (exit 0)

Also ran the full CI-equivalent gate:

```
prek run --from-ref origin/main --to-ref HEAD
```

→ every hook **Passed** (check json, merge conflicts, case conflicts, large files, end of files,
trailing whitespace, TypeScript Check, Oxlint, Oxfmt, i18n Check), exit 0.

## Tests — including reverse-verification

New coverage:

- `tests/unit/media/imagePickError.test.ts` — 7 pure tests on the classifier, including one that
  pins that a plain `Error` whose *message* reads "Path is outside the allowed sandbox." still gets
  the generic text, guarding against a message-sniffing heuristic creeping back in.
- `tests/unit/theme/CssThemeModalCover.dom.test.tsx` — 5 DOM tests for the reported picker.
- Extended `tests/unit/assistants/AssistantEditorSections.dom.test.tsx` (+4) and
  `tests/unit/feedback/InlineAgentEditorAvatar.dom.test.tsx` (+2).

**Reverse-verified.** I reverted only the three call sites to their `origin/main` content, kept the
new tests, and re-ran them:

```
Test Files  3 failed (3)
     Tests  8 failed | 31 passed (39)
```

The 8 failures are exactly the defect: `expected "vi.fn()" to be called 1 times, but got 0 times`
for CssThemeModal (×3) and AssistantEditorSections (×3) — the toast never fires — plus
InlineAgentEditor's `expected "vi.fn()" to be called with arguments: [ 'settings.imagePickOutsideSandbox' ]`
(old code showed the generic message where the specific one applies) and its silent empty-payload
branch. The happy-path and dialog-dismissed cases passed both before and after, as they should.

Restoring the fix: **46/46 pass**.

## Runtime Verification

- [ ] Verified on macOS
- [ ] Verified on Windows
- [ ] Verified on Linux

Not launched as a packaged app — the change is covered by DOM tests that drive the real components
through the actual failing paths. The original report is Windows 11; the sandbox check is
platform-independent, so the fix is too.

## Additional Context

`settings.imagePickOutsideSandbox` was added to **all 13 locales** and `i18n-keys.d.ts` regenerated
(it is a strict key union, so a missing key would fail typecheck). The helper lives in
`components/media/` — the repo's designated file-preview/image domain, which already hosts the other
`getImageBase64` consumers and a precedent non-component module (`webviewHistory.ts`) — keeping that
directory at the 10-child limit rather than pushing an already-full one over it.

### DIFF
--- a/packages/desktop/src/renderer/components/media/imagePickError.ts
+++ b/packages/desktop/src/renderer/components/media/imagePickError.ts
@@ -1,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2025 AionUi (aionui.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isBackendHttpError } from '@/common/adapter/httpBridge';
+import type { TFunction } from 'i18next';
+
+/**
+ * 后端沙箱拒绝的错误码 / Backend error code for a sandbox rejection.
+ *
+ * `POST /api/fs/image-base64` answers 403 with this machine-readable `code`
+ * whenever the requested path resolves outside the allowed roots (temp dir /
+ * home dir / work dir). The native file dialog is free to hand back exactly
+ * such a path — another drive, or a cloud-redirected Desktop — so this is the
+ * most likely way an otherwise valid image fails to load.
+ *
+ * The same code is already consumed by `renderer/utils/previewError.ts`.
+ */
+const PATH_OUTSIDE_SANDBOX = 'PATH_OUTSIDE_SANDBOX';
+
+/**
+ * 把选图失败映射为提示文案 / Map an image-pick failure onto a display message.
+ *
+ * `error` is whatever the picker path produced: a `BackendHttpError` thrown by
+ * the HTTP bridge, some other thrown value, or `null` when the call resolved
+ * but handed back an empty payload.
+ *
+ * Only the sandbox rejection gets a specific, actionable message, because it is
+ * the only case the backend guarantees a stable `code` for. Everything else
+ * falls back to the generic failure text on purpose — matching against error
+ * message strings would be guesswork, and a wrong-but-specific message is worse
+ * than an honest generic one.
+ */
+export const getImagePickErrorMessage = (error: unknown, t: TFunction): string => {
+  if (isBackendHttpError(error) && error.code === PATH_OUTSIDE_SANDBOX) {
+    return t('settings.imagePickOutsideSandbox', {
+      defaultValue: 'This location is not accessible. Pick an image under your home directory or workspace.',
+    });
+  }
+  return t('common.failed', { defaultValue: 'Failed' });
+};
--- a/packages/desktop/src/renderer/pages/settings/AgentSettings/InlineAgentEditor.tsx
+++ b/packages/desktop/src/renderer/pages/settings/AgentSettings/InlineAgentEditor.tsx
@@ -13,6 +13,7 @@ import { resolveAssistantAvatar } from '@/renderer/utils/model/assistantAvatar';
 import { Alert, Avatar, Button, Collapse, Input, Message, Typography } from '@arco-design/web-react';
 import { CheckOne, CloseOne } from '@icon-park/react';
 import EmojiPicker from '@/renderer/components/chat/EmojiPicker';
+import { getImagePickErrorMessage } from '@/renderer/components/media/imagePickError';
 import CodeMirror from '@uiw/react-codemirror';
 import { json } from '@codemirror/lang-json';
 import { useThemeContext } from '@/renderer/hooks/context/ThemeContext';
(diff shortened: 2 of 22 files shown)


kaizhou-lab:
Closing this — the approach is wrong.

The real problem reported in #4073 is that the image cannot be selected at all. This PR only made the failure visible instead of making the selection work, which does not solve the user's problem. The fix should follow however the assistant avatar picker already handles images located outside the backend's allowed roots (temp / home / work dir), rather than surfacing a sandbox rejection as a message.

Reworking from that direction.

PR #4135, opened 2026-08-21
feat(renderer): render WaveDrom timing diagrams in markdown

heartacker:
## Description

Render WaveDrom / WaveJSON timing diagrams (` ```wavedrom ` and ` ```wavejson ` fences) directly in markdown, with the same interaction model as Mermaid blocks:

- **Rendering**: WaveDrom is bundled (v3.6.2, typed via a local declaration module). Valid WaveJSON renders to a responsive inline SVG; invalid input falls back to the source view with JSON5-lenient parsing (comments, trailing commas).
- **Click-to-zoom**: clicking a diagram opens a fullscreen zoom overlay (shared `DiagramZoomOverlay` with Mermaid) with wheel/button zoom, drag-pan, fit-on-open and ESC/backdrop close.
- **Pan & zoom inline** (chat messages and preview panel): drag-to-pan and zoom/reset toolbar buttons for long waveforms.
- **Theme handling**: a `WaveThemeMode` option (`'auto'` follows the app theme with the dark skin, `'light'` always uses the bundled light skin) is **hardcoded to `'light'`** — light diagrams (dark strokes on the light backdrop) stay fully readable on any app theme, and the skin/backdrop pair is derived from a single resolved render theme so they can never drift apart. Flip `WAVEDROM_THEME_MODE` to `'auto'` (or wire it to a setting) to restore theme-following.
- i18n keys added for the toolbar/zoom actions across all 13 locales.

The branch contains 6 commits: the initial support, the zoom overlay, and three follow-up fixes (dark-skin fill remap, overlay card sizing, light-only default).

## Related Issues

N/A

## Type of Change

- [x] This PR contains **exactly one** feature or bug fix that cannot be further decomposed — WaveDrom diagram support for markdown (the follow-up commits are iterations of the same feature)

## Local Checks (Rule 3)

- [x] `bun run lint` — no lint errors (skip if no `.ts`/`.tsx` changed)
- [x] `bunx tsc --noEmit` — no type errors (skip if no `.ts`/`.tsx` changed)
- [x] `bunx vitest run` — tests pass (480 files / 4399 tests, plus a full re-run after a flaky test)
- [x] i18n validated (`bun run i18n:types` + `node scripts/check-i18n.js`) — only if `src/renderer/`, `locales/`, or `src/common/config/i18n/` changed; N/A otherwise

## Runtime Verification

- [ ] Verified on macOS
- [ ] Verified on Windows
- [x] Verified on Linux — diagrams rendered and verified against the built CSS in headless Chrome (theme pairing, overlay sizing, mermaid comparison)

## Screenshots

No screenshots attached — the rendering was verified programmatically (pixel analysis + computed styles in headless Chrome).

## Additional Context

Why light-only by default: the bundled dark WaveDrom skin paints pure-white strokes, which vanish if the backdrop ever resolves to a light value (the token fallback issue that prompted the last fix). Light diagrams are legible on every app theme, so the option defaults to `'light'`; `resolveWaveRenderTheme` is exported for a future setting.


heartacker:
image

heartacker:
image

tcp404:
Nice feature — clean reuse of the Mermaid interaction model, and the test coverage is thorough. CI is green across all platforms. No blocking issues from my read; a few improvement notes and residual risks below.

## Suggestions

**1. Side effect inside `useMemo` (`diagramIndex++`)** — `renderWaveSvg` mutates the module-level `diagramIndex` while being called from `useMemo`. Under StrictMode (dev double-invoke) or concurrent re-render/discard, the index can skip. It only backs the unique `svgcontent_<n>` root id, so uniqueness still holds and there's no functional impact — but it's an impure memo. Consider allocating the id outside the memo (a `useRef` bumped in commit, or `useId`).

**2. A large dark-theme code path is currently unreachable** — with `WAVEDROM_THEME_MODE` hardcoded to `'light'`, `renderTheme` is always `'light'`, so `DARK_SKIN_FILL_REMAP` / `waveSkinDarkRemapped` / `PANEL_BG.dark` / the `auto` branch of `resolveWaveRenderTheme` never execute, and the `MutationObserver` only ends up affecting the source-view highlight. I understand this is a deliberate trade-off (the PR body explains it well). Two small concerns worth noting for the future maintainer: (a) `waveSkinDarkRemapped` depends on the wavedrom skin tree shape (`original[2]` being the `style` element) and silently falls back if that shape changes on a version bump — and because the path is unreachable today, that regression wouldn't be noticed; (b) it's a fair bit of surface area with zero runtime payoff right now. Fine to keep as-is, but consider either trimming the dark branch until the setting is actually wired, or flagging the "prepared / unreachable" nature more prominently.

## Residual risks (not verified in this review)

- **Multi-diagram SVG id collisions**: the incrementing index guarantees a unique root `svgcontent_<n>`, but if the skin `<defs>` carry fixed ids (gradients/markers/glyphs), several WaveDrom diagrams on the same message could collide on `<use>`/referenced ids. Worth a manual check with 2+ diagrams in one message.
- **`dangerouslySetInnerHTML` XSS surface**: WaveJSON comes from markdown (LLM/user) content. Text should be escaped by `onml.stringify` and this matches the existing Mermaid pattern, so the risk looks low — just calling it out.
- **vendor chunk packaging**: `wavedrom` is added to the `manualChunks` `vendor` group. There's prior history in this repo of a vendor-chunk cycle causing a production/WebUI white screen that dev builds don't catch. `wavedrom` doesn't depend on React so a cycle is unlikely, but a `bun run package` sanity check (incl. the WebUI/mobile path) before merge would be reassuring.

## Nits
- `summary`/`previewTitle` re-split `code` on every render — could be memoized.
- Runtime Verification only checked Linux; the tests all mock `renderAny`/`onml.stringify`, so there's no integration test exercising a real wavedrom render (which is where the id-collision / real-SVG behavior would surface).

Overall: looks good to merge after considering the `useMemo` purity nit and the dark-path trade-off. Thanks for the well-documented PR!

heartacker:
In reply to @tcp404's review comment:

Follow-up commit pushed: `cf400e489 fix(renderer): keep WaveDrom diagram ids pure and unique per instance` (branch rebased onto the latest `main`).

**useMemo impurity** — fixed. `renderWaveSvg` no longer bumps a module counter; it takes the index as a parameter. Each instance reserves its diagram id once at mount from a module-level counter (read in the memo, advanced only in the ref initializer / commit effect), so the memo is pure, StrictMode double-invokes merely skip numbers, and — importantly — several instances mounting in the same commit still get distinct `svgcontent_<n>` root ids. `summary`/`previewTitle` are memoized too.

**Integration test** — added `tests/unit/renderer/wavedromBlock.integration.dom.test.tsx`, which renders the **real** wavedrom (no module mocks): two diagrams in one document with unique root ids, real skin geometry (`.s1` wave paths, lane labels, the skin socket rect 20×20/x6/y15 from `laneParamsFromSkin` at index 0), and the light skin's black strokes (`stroke:#000`).

**Multi-diagram id collisions** — verified against the bundled skins: both skins carry ~200 fixed marker/glyph ids in their `defs` (`0m0`, `arrowhead`, `vmv-3-7`, …). Every diagram on a page shares one skin, so the defs are byte-identical; `url(#0m0)` resolves to the first match, which is the same content. Only the `svgcontent_<n>` root ids had to stay unique, which the per-instance allocation now guarantees (previously the in-memo counter made them unique too, but impurely).

**Dark path** — kept as-is per your "fine to keep": light diagrams are legible on every app theme, and `WAVEDROM_THEME_MODE = 'light'` is the single switch point (flip to `'auto'` or wire it to a setting to restore theme-following).

**Vendor chunk** — `bun run package` succeeds (electron-vite build, exit 0); wavedrom is bundled into the vendor chunk alongside the other diagram libs.

Full `bunx vitest run`: 4780 passed / 1 failed — the failure is a pre-existing flaky 10s timeout in `OfficeWatchViewer.dom.test.tsx`'s dynamic `import()` under full-suite parallel load; it passes 14/14 in isolation and is untouched by this PR.

tcp404 (commented):
Thanks for the quick follow-up — the fix looks solid. `renderWaveSvg` is pure now (index passed in), the per-instance id reservation hands out distinct `svgcontent_<n>` roots even for same-commit mounts, and the new real-render integration test (`Set(ids).size === 2` plus the third-mount check) locks exactly the behavior that was at risk. CI is green on all three platforms.

Two small non-blocking notes on the id allocation — happy to merge once these are addressed (or leave them if you prefer; they don't affect behavior):

**1. `useRef(nextDiagramIndex++)` still runs the side effect on every render, not just at mount.** `useRef`'s `initialValue` is a plain function argument, so JS evaluates `nextDiagramIndex++` on every render even though the returned ref keeps its first value — the exact footgun the React docs warn about ("avoid recreating the ref's initial value"). Uniqueness is unaffected since `idRef.current` is locked after the first render, but the counter still increments on every re-render (e.g. every `pointermove` during a pan bumps it), so the "advanced only in the ref initializer at mount" comment isn't quite accurate. The clean form is the lazy init the docs recommend:

```ts
const idRef = useRef<number>(-1);
if (idRef.current < 0) idRef.current = nextDiagramIndex++;
```

**2. The commit effect `useEffect(() => { nextDiagramIndex += 1; }, [svg])` is redundant.** Uniqueness is already guaranteed by the synchronous, ordered ref-init (A runs before B in the same commit, so they get distinct numbers). The extra bump contributes nothing to uniqueness and re-bumps whenever `svg` changes. With the lazy init above you can drop this effect entirely.

Neither affects correctness — LGTM to merge once you've had a look. Thanks again for the well-tested PR!

heartacker:
Thanks for the close read — both notes addressed in `d8d62d722` (branch also rebased onto the latest `main`, tip `573927d46`):

**1. Lazy reservation, no per-render side effect** — `useRef(nextDiagramIndex++)` is gone; the instance now uses the React-documented lazy form:

```ts
const idRef = useRef<number>(-1);
if (idRef.current < 0) {
  idRef.current = nextDiagramIndex++;
}
```

The counter only advances on the first render of each instance; re-renders (pan `pointermove`, theme effects, view-mode switches) no longer consume numbers.

**2. Commit effect dropped** — removed `useEffect(() => { nextDiagramIndex += 1; }, [svg])` entirely. Ordered first-render reservations already give every same-commit mount a distinct number (component A's guard runs before B's), so the bump was pure drift; StrictMode double-renders now don't even skip numbers.

Uniqueness contract unchanged and still locked by the integration test (`Set(ids).size === 2` for same-commit mounts, third mount picks a fresh id). `tsc --noEmit` clean; wavedrom/diagram suites 35/35; full `bunx vitest run` 4839 passed / 0 failed.

PR #4144, opened 2026-08-22
feat(conversation): add Auto planner/worker foundation for aionrs

JeffC0628:
## Summary
- Add product-level **Auto** for aionrs: planner / worker / utility slot settings, sticky Auto in selectors (desktop + mobile + Guid), and `conversation.extra.auto_model`.
- Phase 1 resolves a **concrete** worker-slot model into `conversation.model` (never sends the Auto sentinel to Core). Full long-run cost savings need Phase 2 (switch model without agent rebuild).
- Includes PRD, i18n (13 locales), and unit tests for resolve heuristics.

Closes #4143

## Scope / non-goals
- Does **not** change assistant `mode: auto` (Remember last used).
- Does **not** force Auto as the default model.
- ACP platforms unchanged.

## Test plan
- [ ] Configure Auto slots under Settings → Model; verify persistence across restart
- [ ] In aionrs chat, pick Auto → pill shows `Auto · {model}`; `extra.auto_model.enabled` true
- [ ] Pick a concrete model → Auto flag clears
- [ ] Guid new chat with Auto → create stores `auto_model` on the conversation
- [ ] Mobile action sheet shows Auto and selects correctly
- [ ] `bun run test -- tests/unit/renderer/autoModel/resolveAutoModel.test.ts`
- [ ] `bun run i18n:types` + `node scripts/check-i18n.js`

Issue #4143, opened 2026-08-22
[Feature]: Auto planner/worker model routing for long-running aionrs agents


JeffC0628:
### Feature Description

Add an Auto mode for aionrs aimed at long-running / unattended agents (multi-hour or 24×7-style jobs).

Problem to solve: today a long agent run usually stays on one expensive frontier model for every turn—including low-leverage steps (acks, simple tool follow-ups, mechanical edits, status checks). That wastes money and makes users reluctant to leave agents running.

Desired behavior under Auto:
- High-value work (planning, replanning, hard architecture / tradeoff decisions) → a strong “Planner” model
- Low-value / high-volume work (executing clear steps, routine tool use, mechanical edits) → a cheaper “Worker” model
- Optionally, utility/subagent work (search, summarize, explore) → an even cheaper “Utility” model
- UI shows which model ran, e.g. Auto · worker/<model> or Auto · planner/<model>
- Users can configure which concrete models fill Planner / Worker / Utility slots

This follows the industry planner–worker pattern (Aider Architect/Editor, Claude Code opusplan + per-subagent models, Cursor agent swarm economics, OpenSquilla tier routing)—adapted to AionUi’s BYOK providers and aionrs harness.

### Problem Statement

Long-horizon aionrs tasks spend most turns on low-leverage work. If the session is pinned to a frontier model, every “ok, continue”, tool-result parse, and mechanical edit bills like planning.

Consequences:
1) Unnecessary premium spend and rate-limit pressure
2) Anxiety running overnight / 24×7 jobs (“is this burning frontier tokens on trivia?”)
3) Users either overpay, constantly babysit /model switches, or avoid long unattended runs

We need Auto that routes by value of the step—not just a sticky default in the model picker.

Constraint to call out: changing conversation.model today stops and rebuilds the aionrs agent. A real Auto needs harness support to use different upstream models across turns/phases without tearing down the session (or equivalent: worker turns/subagents on a different model). UI-only hot-swap is not enough and can hurt stability.

Also: assistant defaults.model.mode = “auto” means Remember last used—not cost-aware routing. Keep those concepts separate.

### Proposed Solution

Recommended product shape (best fit for the problem + mature industry pattern):

Auto = configurable Planner + Worker (+ optional Utility), with phase-based switching:
- Planner phase: initial plan, replan after failures, hard decisions
- Worker phase: execute planned steps / routine tool loops
- Utility (optional): explore/search/summarize-style sub-work
[...]

PR #4146, opened 2026-08-22
feat(conversation): Auto phase routing with aionrs model hot-swap

JeffC0628:
## Summary
- Phase 2 for #4143: before each Auto send, decide planner vs worker, resolve the slot model, and hot-swap via `config-options/model` when the provider matches.
- Pill shows `Auto · {phase}/{model}`; Guid first create seeds planner.
- Depends on AionCore hot-swap: https://github.com/iOfficeAI/AionCore/pull/923 (graceful fallback if Core lacks the option).

## Test plan
- [ ] With Core #923: Auto conversation first message uses planner slot; follow-ups use worker without agent rebuild
- [ ] Replan-style user text escalates to planner
- [ ] Cross-provider slot bindings still rebuild (expected)
- [ ] `bun run test -- tests/unit/renderer/autoModel`
- [ ] Without Core #923: send still works (hot-swap fails soft, persist-only)

Related: Phase 1 PR https://github.com/iOfficeAI/AionUi/pull/4144


IceyLiu:
Thanks for this — the direction is right and the engineering hygiene is good (self-contained `autoModel/` module, PRD included, all 13 locales complete, unit tests present).

I ran the full gate locally on `feat/auto-phase2-routing`:

| Check | Result |
| --- | --- |
| `bunx tsc --noEmit` | ✅ 0 errors |
| `bun run test -- tests/unit/renderer/autoModel` | ✅ 9/9 passed |
| `bun run i18n:types` + `scripts/check-i18n.js` | ✅ in sync, all locales |
| `bun run lint --quiet` | ✅ 0 errors |

Note for reviewers: the `feat/auto-phase2-routing` branch already contains all commits from #4144, so merging this PR covers both.

Below are the blocking and should-fix items before merge.

---

## 🔴 Blocking: unmerged AionCore dependency + a regression on the existing model-switch path

This PR depends on [AionCore#923](https://github.com/iOfficeAI/AionCore/pull/923), which is still open. The Auto path itself degrades gracefully (`setConfigOption` is wrapped in try/catch), so that part is fine.

The problem is `ChatConversation.tsx` — `onSelectModel` no longer stops the active turn when the provider is unchanged:

```ts
const sameProvider = conversation.model?.id === selected.id;
if (runtimeView.activeTurnId && !sameProvider) { /* stop */ }
```

This changes behaviour for **all** manual model switches, not just Auto. Without Core#923, Core does not hot-swap either, so a user switching models mid-turn within the same provider gets a UI that says the model changed while the turn keeps running on the old one — a silent no-op regression on a path everyone uses.

Please either land and release Core#923 first, or split this `sameProvider` stop-skip out of the PR so it ships together with the Core support.

## 🟠 Vision filtering is dead code — image sends can pick a text-only model

`requireVision` is declared in `ResolveAutoModelInput` and honoured inside `resolveAutoModel`, but no caller ever passes it:

```
resolveAutoModel.ts:34   if (input.requireVision && ...)
resolveAutoModel.ts:53   if (input.requireVision && ...)
types.ts:48              requireVision?: boolean;
```

`AionrsSendBox` has `files` in scope at the `applyAutoModelForTurn` call site, and `handleSelectAuto` has no vision context at all. Sending an image in an Auto conversation can therefore resolve a non-vision worker model and fail the request. Please thread `requireVision` through from the send path (and consider re-resolving when attachments are present).

## 🟠 The tier heuristics rank the wrong models, and the preference switch is a no-op

I ran `scoreModelForSlots` / `preferenceWeight` over a realistic catalogue (`claude-opus-4-8`, `claude-sonnet-4-6`, `claude-haiku-4-5`, `gpt-5`, `gpt-4o-mini`, `deepseek-chat`, `kimi-k2`, `glm-4.6`, `qwen3-coder-plus`, `gemini-3-pro-preview`):

```
cost/planner    -> gemini-3-pro-preview:5   opus:4  gpt-5:4  sonnet:3
balance/planner -> gemini-3-pro-preview:7   opus:4  gpt-5:4  sonnet:3
quality/planner -> gemini-3-pro-preview:14  opus:8  gpt-5:8  sonnet:6

cost/worker     -> qwen3-coder-plus:12  gemini-3-pro-preview:10  sonnet:8  haiku:8
balance/worker  -> qwen3-coder-plus:12  sonnet:8  deepseek-chat:8
quality/worker  -> qwen3-coder-plus:12  sonnet:8.9  gemini-3-pro-preview:8.1

cost/utility    -> haiku:6  gpt-4o-mini:6  gemini-3-pro-preview:6
balance/utility -> haiku:4  gpt-4o-mini:4  gemini-3-pro-preview:4
```

Four concrete problems:

1. **`gemini-3-pro-preview` always wins `planner`**, ahead of Opus and GPT-5, because it double-scores on both `/pro-preview/i` (+4) and `/gemini.*pro/i` (+3) while `/opus/i` only yields +4.
2. **`qwen3-coder-plus` always wins `worker`** (12) by double-scoring `/qwen.*plus/i` and `/coder/i`. Any model with `coder` in its name captures the worker seat regardless of size or price.
3. **`cost` / `balance` / `quality` produce identical winners in every seat.** The preference selector in Settings has no observable effect on this catalogue.
4. `tierHeuristics.ts:41` is an identity expression:
   ```ts
   utility: Math.max(utility, planner > 6 ? 0 : utility)
   ```
   `Math.max(u, 0) === u` for all `u >= 0`, and `Math.max(u, u) === u`. The intended "zero out utility for strong planners" never happens — which is why `gemini-3-pro-preview` also ties for the `utility` seat.

Also, `kimi-k2` and `glm-4.6` score `{planner:0, worker:4, utility:0}` — treated as unknown. Regex-guessing model names is going to keep drifting as new models ship. Consider an explicit capability/tier table, or requiring explicit slot bindings and treating `automatic` as a last resort.

## 🟠 `hasPriorUserTurns` resets on every conversation open, so reopening re-runs planner

```ts
useEffect(() => { autoUserTurnsRef.current = 0; }, [conversation_id]);
...
hasPriorUserTurns: autoUserTurnsRef.current > 0
```

The counter is in-memory only. Reopen a 50-turn Auto conversation after a restart and the next message is classified as the first turn → planner. That defeats the cost-saving goal in the most common long-running scenario, which is exactly the one #4143 targets. Derive this from message history or from the persisted `extra.auto_model` state instead.

## 🟡 Two features are wired to nothing

- `consecutiveWorkerFailures` is read in `decideAutoModelPhase` and forwarded by `applyAutoModelForTurn`, but no production caller ever supplies it. The "≥2 worker failures escalate to planner" rule in the PRD is unreachable outside the unit test.
- The `utility` slot is configurable in Settings but never consumed by any resolve path (PRD defers it to F-AUTO-07). Users can bind a model to a seat that does nothing — please hide it until it is wired, or wire it.

## 🟡 `AutoModelSettingsBlock` renders above the empty state

In `ModelModalContent.tsx` the block is now rendered unconditionally at the top of the scroll area, ahead of the `!data || data.length === 0` branch. A first-run user with zero providers sees three empty slot dropdowns before the "no configured models" onboarding copy. Suggest moving it below the provider list and gating it on `data?.length`.

## 🟡 Auto is silently skipped for team conversations

The send path is guarded by `if (autoEnabled && !teamSendMessage)`, but the pill still renders `Auto · {phase}/{model}`. The UI claims routing that isn't happening.

## 🟡 Per-turn latency: 3 extra round trips before each send

`useAcpConfigOptions.setConfigOption` does `prepareSetRuntime()` → `fetchConfigOptionsOnce()` → `PUT config-options/model`, and then `applyAutoModelForTurn` issues a `PATCH /conversations/:id`. That's 3–4 sequential round trips added in front of every Auto send, on the critical path to first token. Worth short-circuiting when the resolved model equals the current one (the `alreadyOnModel` check covers the PUT but not the PATCH), and consider skipping the pre-fetch.

## 🟡 Possible bypass of a locked `model` option

`handleSheetModelSelect` guards on `runtimeConfig.isConfigOptionBlocked?.('model')`, but the Auto send path does not. If `setConfigOption` throws `config_update_in_progress` (or the option is blocked by an assistant that pins the model), the error is swallowed by the try/catch and the code still persists `conversation.model`. Please confirm Auto respects the same block.

---

## Housekeeping

The branch is 12 commits behind `main` — please rebase so the merge diff can be re-checked.

---

Happy to re-review once the Core dependency is resolved and the vision / first-turn / heuristics items are addressed. The module boundaries and test setup are solid, so most of this is contained work.

JeffC0628:
Addressed @IceyLiu review feedback in latest push (rebased on main):

- **Blocking**: Restored stop-on-model-change for manual model switches during active turns (same-provider hot-swap only on Auto per-turn path).
- **Should-fix**: \
equireVision\ wired when send includes image attachments; tier heuristics/preference scoring fixed; Auto settings hidden when no providers configured.

Superseded #4144 (closed). Auto hot-swap at send time still depends on AionCore #923.

JeffC0628:
Addressed @IceyLiu review feedback in latest push (rebased on main):

- **Blocking**: Restored stop-on-model-change for manual model switches during active turns (same-provider hot-swap only on Auto per-turn path).
- **Should-fix**: \
equireVision\ wired when send includes image attachments; tier heuristics/preference scoring fixed; Auto settings hidden when no providers configured.

Superseded #4144 (closed). Auto hot-swap at send time still depends on AionCore #923.

JeffC0628:
Addressed @IceyLiu review feedback in latest push (rebased on main):

- **Blocking**: Restored stop-on-model-change for manual model switches during active turns (same-provider hot-swap only on Auto per-turn path).
- **Should-fix**: \
equireVision\ wired when send includes image attachments; tier heuristics/preference scoring fixed; Auto settings hidden when no providers configured.

Superseded #4144 (closed). Auto hot-swap at send time still depends on AionCore #923.

JeffC0628:
Addressed @IceyLiu review feedback (rebased on main, CI green):

- **Blocking**: Restored stop-on-model-change for manual model switches during active turns.
- **Should-fix**: requireVision when send includes images; tier heuristics/preference scoring fixed; Auto settings hidden when no providers configured.

#4144 closed (superseded). Auto per-turn hot-swap still depends on AionCore #923.

JeffC0628:
Additional hardening after self-review:

- **Fix**: \hasPriorUserTurns\ now derives from loaded message history (\conversationHasUserTurns\) instead of a session-local counter — reopening an existing Auto conversation no longer mis-routes every send to planner.
- **Tests**: added \pplyAutoModelForTurn\, vision/history helpers, requireVision filtering, and balance worker selection coverage (22 autoModel tests total).

Issue #4155, opened 2026-08-23
WebUI: paired browser dies ~24h after pairing — endless 401 + WebSocket reconnect loop


ajdevy:
### Module

WebUI & Remote Access

### Platform

macOS (Apple Silicon / Intel)

### AionUi Version

2.1.59

## Bug Description

A phone paired to the desktop WebUI becomes completely unusable roughly **24 hours after pairing**. The page looks like it is reloading about once a second and no interaction is possible. Only re-pairing recovers it — and then it breaks again a day later.

**Heads-up for triage:** the primary defect is in the packaged build, not in this repository. I checked out `main` and grepped for the desktop-pairing WebUI auth code — `desktopPairing`, `createPairedSessionCookies`, `coreUserBridge`, `AuthSessionStore`, `aionui-webhost-session`, `auth-sessions.json`, `reportCoreSessionUnauthorized`, `scheduleSessionRenewal` — **zero hits for all of them**. Public `packages/web-host/src/static-server.ts` is 285 lines with no auth at all and `WebHostOptions` has no `auth` field, while the shipped `@aionui/web-host` in `app.asar` is 802 lines and includes the whole pairing system. So the root cause below can only be fixed in the closed-source build.

One *contributing* defect **is** in this repo, and it is the one that produces the visible "constantly reloading" symptom. PR follows.

## Root cause

Four things compose into a permanent zombie state.

**1. A paired browser's Core session is never renewed.** `createPairedSessionCookies()` mints a Core session JWT once, in the `/auth/pair` handler, and hands it to the browser. The desktop renews its *own* session — the log line is `[CoreUserBridge] session renewal scheduled in 1380min` (23 h; with `RENEWAL_MARGIN_MS = 1h` that puts the JWT TTL at exactly 24 h). Nothing ever schedules a renewal for paired browsers.

**2. `/auth/status` does not reflect Core session validity.** In `desktopPairing` mode it checks only the web-host pairing session (30-day TTL) plus `getDesktopStatus()`. At T+24 h it still answers `{"authenticated": true, "user": {...}}`, so the SPA boots as signed-in and never routes to the pairing screen. Verified directly against the running host:

```
$ curl -s localhost:25808/auth/status -H "Cookie: aionui-webhost-session=<id>"
{"mode":"desktopPairing","authenticated":true,"user":{...}}

[...]

PR #4156, opened 2026-08-23
fix(webui): stop bridge websocket reconnect storm on repeated failures

ajdevy:
# Pull Request

## Description

The WebUI browser bridge reconnects its WebSocket on demand so that a login can immediately get a cookie-bearing connection. That on-demand path bypassed the backoff, which turns any persistent connection failure into a reconnect storm.

Three related defects in `packages/desktop/src/common/adapter/browser.ts`, all on the same code path:

1. **`ensureSocket()` bypassed the backoff.** It called `connect()` directly whenever the socket was `CLOSED`/`CLOSING`, and it is invoked from `emit()` — i.e. on *every* bridge event. With a pending `reconnectTimer` sitting unused, the reconnect rate tracked how often the app emitted rather than the 500 ms→8 s schedule.

2. **`ensureSocket()` ignored `shouldReconnect`.** After `isRealtimeAuthTerminalError()` sets `shouldReconnect = false`, clears the timer and closes the socket, the very next `emit()` re-dialled anyway — so "stopping reconnection" never actually stopped.

3. **The backoff reset on `open`.** A server that completes the WebSocket handshake and then drops the connection (rejected auth, backend restarting) still fires `open`, so `reconnectDelay` snapped back to 500 ms on every attempt and never grew.

Together these mean a WebUI whose connection keeps failing reconnects roughly once a second, forever. Because a reconnect also triggers a full data revalidation, the page appears to be constantly reloading and becomes unusable.

### Changes

- `ensureSocket()` returns early when reconnection is disabled or a retry is already queued.
- `reconnectDelay` only resets after a connection has held for `STABLE_CONNECTION_MS` (5 s), tracked via `connectedAt` on `open` and evaluated on `close`.
- `__websocketReconnect()` clears any pending timer before dialling, so the login flow still reconnects instantly.
- Extracted `BASE_RECONNECT_DELAY` / `MAX_RECONNECT_DELAY` so the two previously duplicated `500` literals and the `8000` cap stay in sync.

Message queueing is unchanged: events emitted while disconnected still land in `messageQueue` and flush on the next `open`. There is a test covering that so the fix cannot silently start dropping events.

## Related Issues

Contributing cause of #4155. It does **not** close that issue — the primary defect there (a paired browser's Core session is minted with a 24 h TTL and never renewed) lives in code that is not part of this repository. This PR fixes the part that is here: the reconnect storm that turns that auth failure into an unusable, seemingly-reloading page.

- Related to #4155

## Type of Change

- [ ] `feat` — New feature (non-breaking change which adds functionality)

The three changes share one root cause and one code path — reconnect scheduling in the browser bridge. Splitting them would leave the storm in place: guarding `ensureSocket()` alone still lets an accept-then-drop server hammer at the 500 ms floor, and fixing only the backoff reset still lets `emit()` skip the timer.

## Local Checks (Rule 3)

- [x] `bun run lint` — no new warnings from the changed files (0 errors; the 4 pre-existing warnings in `browser.ts` are unused catch params in untouched code)
- [x] `bunx vitest run` — 4822 passed, 5 skipped, 0 failed (508 files)
- [x] i18n validated — N/A, no `src/renderer/`, `locales/` or `src/common/config/i18n/` changes
- [x] New/changed user-facing text uses i18n keys (no hardcoded strings) — N/A, no user-facing text

## Runtime Verification

- [x] Verified on macOS
- [ ] Verified on Windows
- [ ] Verified on Linux

Found on macOS 15.6 with a phone paired over LAN to the desktop WebUI, against a backend rejecting the browser's expired session — the exact `accept upgrade, then drop` case in defect 3. Backend logs showed `path=/ws status=101` about once per second alongside a full `/api/*` burst on each reconnect.

New test file `tests/unit/renderer/webuiWebSocketReconnect.dom.test.ts` (7 tests) drives the real module in jsdom with fake timers and a scripted `WebSocket` stub. Against the unpatched adapter **4 of the 7 fail**, so the tests demonstrably pin the regressions rather than just passing:

```
× does not re-dial on every emit while a backoff reconnect is pending
× queues messages emitted while disconnected and flushes them on reconnect
× backs off when the server accepts the upgrade and drops it immediately
× stops reconnecting after a terminal auth error, including on later emits
```

## Additional Context

Reconnect behaviour after the change, for a connection that keeps failing:

| | before | after |
|---|---|---|
| socket closes, app keeps emitting | one dial per emit (~1 Hz+) | one dial per backoff tick |
| server accepts upgrade then drops | 500 ms forever | 500 ms → 1 s → 2 s → 4 s → 8 s |
| terminal auth error, then an emit | reconnects anyway | stays down |
| login calls `__websocketReconnect()` | immediate | immediate (unchanged) |
| connection recovers and holds ≥5 s | delay already 500 ms | delay resets to 500 ms |

### DIFF
--- a/packages/desktop/src/common/adapter/browser.ts
+++ b/packages/desktop/src/common/adapter/browser.ts
@@ -79,10 +79,19 @@ if (win.electronAPI) {
 
   type QueuedMessage = { name: string; data: unknown };
 
+  const BASE_RECONNECT_DELAY = 500;
+  const MAX_RECONNECT_DELAY = 8000;
+  // A connection must hold this long before the backoff is considered recovered.
+  // A server that accepts the upgrade and drops it right away (rejected auth,
+  // backend restarting) still fires `open`, so resetting the delay there alone
+  // would keep the backoff pinned at its minimum forever.
+  const STABLE_CONNECTION_MS = 5000;
+
   let socket: WebSocket | null = null;
   let emitterRef: { emit: (name: string, data: unknown) => void } | null = null;
   let reconnectTimer: number | null = null;
-  let reconnectDelay = 500;
+  let reconnectDelay = BASE_RECONNECT_DELAY;
+  let connectedAt = 0;
   let shouldReconnect = true; // Flag to control reconnection
 
   const messageQueue: QueuedMessage[] = [];
@@ -109,7 +118,7 @@ if (win.electronAPI) {
 
     reconnectTimer = window.setTimeout(() => {
       reconnectTimer = null;
-      reconnectDelay = Math.min(reconnectDelay * 2, 8000);
+      reconnectDelay = Math.min(reconnectDelay * 2, MAX_RECONNECT_DELAY);
       connect();
     }, reconnectDelay);
   };
@@ -134,7 +143,7 @@ if (win.electronAPI) {
     const currentSocket = socket;
 
     currentSocket.addEventListener('open', () => {
-      reconnectDelay = 500;
+      connectedAt = Date.now();
       flushQueue();
     });
 
@@ -212,6 +221,14 @@ if (win.electronAPI) {
         socket = null;
       }
 
+      // Only credit the backoff if the connection actually held for a while.
+      // Otherwise an accept-then-immediately-drop server keeps the delay at its
+      // minimum and the client hammers it forever.
+      if (connectedAt !== 0 && Date.now() - connectedAt >= STABLE_CONNECTION_MS) {
+        reconnectDelay = BASE_RECONNECT_DELAY;
+      }
+      connectedAt = 0;
+
       scheduleReconnect();
     });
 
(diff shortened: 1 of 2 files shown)


ajdevy:
Pushed `e299e22` to close the patch-coverage gap Codecov flagged.

The 2 missing lines + 1 partial were all the same spot — the timer-clearing branch in `__websocketReconnect()`:

```ts
if (reconnectTimer !== null) {
  window.clearTimeout(reconnectTimer);
  reconnectTimer = null;
}
```

Every existing test reached that function *after* a terminal auth error, and that handler already nulls `reconnectTimer` on its way out — so the branch body never executed.

Added `cancels a pending backoff retry when the login flow reconnects`, which leaves a retry genuinely queued and then signs in. It asserts both halves of what that branch is for: login dials immediately instead of waiting out the backoff, and the superseded timer does not later fire a duplicate connect. All 18 executable lines in the diff are now covered — patch coverage 100%, no partials.

tcp404:
Thanks for the exceptionally thorough report and PR — the root-cause breakdown, the before/after table, and especially the test file that fails 4/7 against the unpatched adapter make this easy to reason about and trust.

Confirmed all three defects on current `main`: `ensureSocket()` re-dials on every emit (bypassing the backoff), ignores `shouldReconnect`, and the backoff resets on `open`.

This composes cleanly with the silent WebUI session-refresh that just landed in #4175. The two are complementary: #4175 removes the common trigger (an expired-but-renewable session now refreshes instead of looping), while this PR hardens the reconnect path for any persistent failure — including when a refresh itself fails. I test-merged onto current `main`: no conflict, `tsc` clean, and your new suite plus the existing browser/adapter tests all pass. Nice detail that `ensureSocket()` now respecting `shouldReconnect` also makes the refresh-failure → stay-down path from #4175 actually stick.

Merging — thanks again for the careful work!

Issue #4007, opened 2026-08-12
Slash commands palette is empty when opening existing ACP conversations


nikitatwo404-prog:
### Module

Conversation & Sessions

### Platform

Windows

### AionUi Version

2.1.53

### Bug Description

When opening an existing ACP conversation (e.g. OMP / custom ACP agent), the slash command palette (`/`) in the chat sendbox shows only built-in AionUi commands (like `Add File`), while agent slash commands (e.g. `/mcp`, `/compact`, `/skill:*`) are missing.

However, when switching the conversation agent to an `aionrs`/`antigravity` agent (or when creating a fresh conversation), all slash commands appear correctly.

### Steps to Reproduce

1. Connect a custom ACP agent (e.g. `omp.exe acp`).
2. Create a new conversation and verify slash commands (`/`) appear.
3. Close/restart AionUi or switch between conversations.
4. Re-open the existing ACP conversation and type `/` in the message input.
5. Observe that only Aion's built-in commands appear, while ACP agent slash commands are missing.

### Expected Behavior

The slash command palette should consistently display the full list of available agent slash commands in existing ACP conversations upon loading.

### Actual Behavior

In existing ACP conversations, the slash command palette defaults to empty/built-ins because `AcpChat` (`AcpChat-D-4koxzR.js`) defines `fetchSlashCommands` inside the `Qn` hook but never invokes it on mount, relying solely on the live `available_commands` stream event which may be missed during `session/load`.

Calling `GET /api/conversations/{id}/slash-commands` directly returns HTTP 200 with all 80+ commands intact, proving the backend cache is valid, but the renderer UI does not pull it.

### Additional Context

- `AionrsChat-D3Tr7KDu.js` correctly triggers `getSlashCommands.invoke` in a `useEffect` on mount, which is why switching to an `aionrs` agent (like Gemini) populates commands immediately.
- Suggested fix: invoke `fetchSlashCommands` on mount inside `AcpChat` / `Qn` hook after session warmup/load resolves.

PR #4186, opened 2026-08-26
fix(acp): populate slash command palette on mount for existing conversations (#4007)

aniruddhaadak80:
## What Problem This Solves

When opening an existing ACP conversation, the slash command palette (`/`) shows only built-in AionUi commands while agent slash commands (`/mcp`, `/compact`, `/skill:*`) are missing. Creating a fresh conversation or switching to an `aionrs` agent shows all commands correctly.

## Why This Change Was Made

`useAcpMessage` defines `fetchSlashCommands` but never invokes it on mount. The hook relies solely on the live `available_commands` stream event, which may be missed during `session/load` for existing conversations. The `aionrs` chat correctly triggers `getSlashCommands.invoke` in a `useEffect` on mount.

## User Impact

Existing ACP conversations now show the full slash command palette immediately on load, matching the behavior of fresh conversations and `aionrs` agents.

## Evidence

- Issue: https://github.com/iOfficeAI/AionUi/issues/4007
- Root cause confirmed at `useAcpMessage.ts:731` — `fetchSlashCommands` defined but never called on mount
- Backend `GET /api/conversations/{id}/slash-commands` returns HTTP 200 with all 80+ commands — the cache is valid
- Fix: added `useEffect(() => { fetchSlashCommands(); }, [fetchSlashCommands])` matching the `aionrs` pattern
- 1 file changed, 6 insertions

### DIFF
--- a/packages/desktop/src/renderer/pages/conversation/platforms/acp/useAcpMessage.ts
+++ b/packages/desktop/src/renderer/pages/conversation/platforms/acp/useAcpMessage.ts
@@ -739,6 +739,12 @@ export const useAcpMessage = (
       .catch(() => {});
   }, [conversation_id, options?.prepareRuntime]);
 
+  // Populate the slash command palette on mount so existing ACP conversations
+  // show agent commands immediately, not only after a live stream event.
+  useEffect(() => {
+    fetchSlashCommands();
+  }, [fetchSlashCommands]);
+
   return {
     thought,
     setThought,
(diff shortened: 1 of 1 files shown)


aniruddhaadak80:
Diagnosis of the red Unit Tests (all 3 OSes, run 32990398650): single failure is `tests/unit/renderer/useAcpMessage.dom.test.ts > uses injected runtime preparation for initial slash commands in team mode` — `AssertionError: expected "vi.fn()" to be called 2 times, but got 3 times`.

Root cause is in this PR's diff: the new mount `useEffect(fetchSlashCommands)` duplicates the existing warmup effect, which already calls `prepareRuntime` and fetches slash commands on mount. So on mount `prepareRuntime` ran twice (warmup + new effect) plus once more on the test's manual `fetchSlashCommands()` = 3x instead of the expected 2x.

Fixed in `ee6f7716`: the mount fetch is now guarded to run only when warmup was skipped (`skipWarmup && !prepareRuntime`), keeping this PR's intent for that path without the duplicate `prepareRuntime` + IPC round-trip on the normal path.

PR #4245, opened 2026-09-11
fix(conversation): always send type and extra in create body (aioncore v0.1.72 compat)

shoppingarena:
## Problem

On aioncore **v0.1.72**, every first submit from the home screen fails with `400 Invalid JSON request body` when a conversation is created with an assistant selected.

## Root cause

The backend now requires the top-level `type` and `extra` fields in `POST /api/conversations`. However:

- `buildCreateConversationBody()` sets `type: undefined` whenever an assistant is present (JSON.stringify then drops the key), and drops `extra` when it is not set
- the `useGuidSend` call sites never pass `type`

Resulting body `{"assistant":{...}}` (no `type`, no `extra`) → 400.

Verified against a live v0.1.72 backend:
- `{"assistant":{"id":X}}` → **400**
- `{"assistant":{"id":X},"type":"aionrs","extra":{}}` → **201**

## Fix

- always pass `type` through in the body builder and default `extra` to `{}`
- set the engine type (`'aionrs' | 'acp'`) at both `useGuidSend` call sites

Verified end-to-end after rebuild: first submit with an assistant creates the conversation successfully.

### DIFF
--- a/packages/desktop/src/common/adapter/apiModelMapper.ts
+++ b/packages/desktop/src/common/adapter/apiModelMapper.ts
@@ -54,13 +54,14 @@ export type CreateConversationBodyInput = {
  * agent types carry model info via `extra`.
  */
 export function buildCreateConversationBody(p: CreateConversationBodyInput): Record<string, unknown> {
-  const hasAssistant = p.assistant !== undefined && p.assistant !== null;
+  // aioncore >= v0.1.72 rejects create bodies without top-level `type` and
+  // `extra` ("Invalid JSON request body"), so both must always be present.
   const body: Record<string, unknown> = {
-    type: hasAssistant ? undefined : p.type,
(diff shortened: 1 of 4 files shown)


shoppingarena:
Updated the 4 unit-test assertions that encoded the old contract (`type` stripped when an assistant is present) — they now expect the engine type to be passed through, per the aioncore v0.1.72 requirement. Both affected test files pass locally (40/40).

For the maintainers: on the v2.1.61 + aioncore v0.1.72 combo this is reproducible live — first submit from the home screen with any assistant selected returns `400 Invalid JSON request body`, because the emitted body `{"assistant":{...}}` has no `type`/`extra`. Happy to adjust if you prefer fixing this on the backend side instead (making those fields optional again).

